PR #2, opened 2020-03-12
<feature> rds minor version support

roleyfoley:
Add the ability for the Instance version to include a minor version indication, initially for MySQL.


roleyfoley:
See https://github.com/codeontap/gen3/pull/1260 For further conversation

PR #20, opened 2020-04-04
<feature>Add Userpool resources and CFN support

roleyfoley:
## Description
AWS implementation of https://github.com/hamlet-io/engine/pull/1301 
Moves all configuration from Cli into cloudformation as it now supports everything we currently have in place for userpools
Enables advanced security services to provide stronger identity management

## Motivation and Context
Server to server cognito authentication is useful in products where a service needs to call an API. This interaction can now use the same userpool that is used for users themselves rather than having to implement another auth method 

Refactor: Cloudformation configuration is easier to understand and manage change

## How Has This Been Tested?
Tested locally with a new deployment 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X ] None of the above.

### DIFF
--- a/aws/components/userpool/id.ftl
+++ b/aws/components/userpool/id.ftl
@@ -35,3 +35,15 @@
             AWS_COGNITO_SERVICE
         ]
 /]
+
+
+[@addResourceGroupInformation
+    type=USERPOOL_RESOURCE_COMPONENT_TYPE
+    attributes=[]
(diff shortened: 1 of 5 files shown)


roleyfoley:
Tested migration for CLI deployed resources and cloudformation deployed resources and all updates went through without a problem. Really surprised that the domain name configuration didn't have a problem. It just said that the resource was created. Yay!

PR #34, opened 2020-05-04
feat: Enhanced checks on userpool auth provider names

ml019:
## Description
For some identity providers on a userpool, the provider name that can be used is fixed so ensure the correct name has been used.

## Motivation and Context
Avoid obscure errors generated when the correct provider name is not used.

## How Has This Been Tested?
Local deployment

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/components/userpool/setup.ftl
+++ b/aws/components/userpool/setup.ftl
@@ -363,6 +363,26 @@
                 }]
             [/#list]
 
+            [#local requiredProviderName = ""]
+            [#switch authProviderEngine]
+                [#case "Facebook"]
+                [#case "Google"]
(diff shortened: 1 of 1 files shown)


ml019:
Interestingly, its not explicitly listed in the fixed values for AuthProviders butit feels like it shoul;d be. I just reflected the fixed ones per the documentation. @roleyfoley reckon I can add for Apple as well as I suspect the same rule will apply and it doesn't hurt?

roleyfoley:
Lets merge it based on whats in the doco, we can look at it later when we have an apple app to test with

PR #45, opened 2020-05-20
feat(globaldb): initial support for the globalDb component

roleyfoley:
## Description
AWS Implementation of https://github.com/hamlet-io/engine/pull/1325 
Supports the provisioning of globalDbs using the AWS DynamoDB service 

## Motivation and Context
Allows for the deployment and control of system level configuration of DyanmoDB Tables 

## How Has This Been Tested?
Tested locally using active deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/globaldb/state.ftl
+++ b/aws/components/globaldb/state.ftl
@@ -1,0 +1,51 @@
+[#ftl]
+
+[#macro aws_globaldb_cf_state occurrence parent={} ]
+    [#local core = occurrence.Core]
+    [#local solution = occurrence.Configuration.Solution]
+
+    [#local id = formatResourceId(AWS_DYNAMODB_TABLE_RESOURCE_TYPE, core.Id )]
+    [#local key = solution.PrimaryKey ]
+
+    [#local sortKey = "" ]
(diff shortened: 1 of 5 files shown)


aws/components/globaldb/state.ftl
```diff
@@ -0,0 +1,51 @@
+[#ftl]
+
+[#macro aws_globaldb_cf_state occurrence parent={} ]
+    [#local core = occurrence.Core]
+    [#local solution = occurrence.Configuration.Solution]
+
+    [#local id = formatResourceId(AWS_DYNAMODB_TABLE_RESOURCE_TYPE, core.Id )]
```
ml019, line 7: Table names all need to include product environment and segment (if not default) to ensure uniqueness. Think this is something we should automatically add.

From a coding perspective, might also be useful to know what this prefix is, though if using linking, then the table name is one of the attributes so they should really care.

roleyfoley: Current implementation was using the randomly generated CFN name, for the globaldb I've added naming support using core.FullName 

Prefix isn't really part of this since we give them the table and it has the full name

PR #47, opened 2020-05-26
Router component

roleyfoley:
## Description
AWS support for https://github.com/hamlet-io/engine/pull/1327 

Adds the router component based on the AWS Transit Gateway service. The router deployment itself is a single transit gateway with a single route table. 

The attachment and routing of networks on the Transit gateway is managed by the gateway service. The gateway service should link to a router and based on this the gateway will attach to the transit gateway, route propagation is enabled by default to reduce the configuration required for the networks. 

Adds support for using AWS Resource Access Manager which allows for sharing specific resources between AWS accounts, in this case allows sharing the router between different accounts to form a network mesh between different AWS account VPCs

## Motivation and Context
Supports routing between different AWS VPCs or on premises networks using VPN or Direct Connect links

## How Has This Been Tested?
Tested on local deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/gateway/setup.ftl
+++ b/aws/components/gateway/setup.ftl
@@ -112,6 +112,16 @@
             [#case "vpcendpoint"]
             [#break]
 
+            [#case "router"]
+                [#local transitGatewayId = ""]
+                [#local routerFound = false]
+
+                [#local transitGatewayAttachementId = gwResources["transitGatewayAttachement"].Id ]
+                [#local transitGatewayAttachementName = gwResources["transitGatewayAttachement"].Name ]
+                [#local transitGatewayRoutePropogationId = gwResources["routePropogation"].Id ]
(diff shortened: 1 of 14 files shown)


aws/components/gateway/setup.ftl
```diff
@@ -149,6 +166,97 @@
...
+
+                [@createTransitGatewayAttachment
+                    id=transitGatewayAttachementId
+                    name=transitGatewayAttachementName
+                    transitGateway=transitGateway
+                    subnets=getReferences(attachementSubnets)
+                    vpc=getReference(vpcId)
+                /]
+
+                [@createTransitGatewayRouteTablePropogation
```
RossMurr4y, line 246: Should this one be Propagation?

roleyfoley: Thanks fixed that up

PR #57, opened 2020-06-05
feat(router): support for static routes

roleyfoley:
## Description
Implements https://github.com/hamlet-io/engine/pull/1331 for AWS 

To implement this you need to create a couple of things 

In the client account:
- An external service with the transitGateway Id set as the TRANSIT_GATEWAY_ID attribute 
- A Gateway using the router engine with a link to this external service 

 This will attach the client VPC to the Transit Gateway 

In the owner account:
- A router component 
- An external service with the transitGateway Attachement from the client Id as TRANSIT_GATEWAY_ATTACHMENT attribute 
- An IP address group which represents the client VPC Network 
- A Static Router on the router which links to the external Service for the transit gateway attachment

## Motivation and Context
Allows for transit gateways shared between two AWS accounts, the account who has received the transit gateway can only create attachments,  routes, routeTables and propagation configuration can only be managed by the account which owns the gateway. 

## How Has This Been Tested?
Tested on local deployment 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/gateway/state.ftl
+++ b/aws/components/gateway/state.ftl
@@ -66,7 +66,7 @@
                         "Name" : core.FullName,
                         "Type" : AWS_VPC_IGW_RESOURCE_TYPE
                     },
-                    "internetGatewayAttachement" : {
+                    "internetGatewayAttachment" : {
                         "Id" : formatId(AWS_VPC_IGW_ATTACHMENT_TYPE, core.Id),
                         "Type" : AWS_VPC_IGW_ATTACHMENT_TYPE
                     }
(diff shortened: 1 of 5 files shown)


aws/components/gateway/state.ftl
```diff
@@ -74,12 +75,15 @@
...
-                        "Id" : formatResourceId(
+
+            [#local transitGatewayAttachementId = formatResourceId(
                             AWS_TRANSITGATEWAY_ATTACHMENT_RESOURCE_TYPE,
                             core.Id
-                        ),
+                        )]
+
+            [#local resources += {
+                "transitGatewayAttachement" : {
```
ml019, line 85: Attachement -> Attachment?

aws/components/gateway/state.ftl
```diff
@@ -74,12 +75,15 @@
...
+                        "Id" : transitGatewayAttachementId,
```
ml019, line 86: Same here -maybe go through and global update?

aws/components/gateway/state.ftl
roleyfoley: fixeded

aws/components/gateway/state.ftl
roleyfoley: fixeded

PR #105, opened 2020-08-02
feat(waf): enable log waf logging for waf enabled services

roleyfoley:
## Description
Implementation of https://github.com/hamlet-io/engine/pull/1366  which enables log forwarding for WAF enabled services

To setup WAF logging you need 
- A datafeed component deployed with a backend configured ( for s3 no lambda transfromation is required, for Elasticsearch this may be required ) 
  - The datafeed also needs to be configured as a WAFLogFeed using the `aws:WAFLogFeed` configuration attribute. This changes the name of the deployed firehose to align with WAF setup requireemnts 
- A logging profile which links to this datafeed component 
   - Assign this logging profile under the WAF configuration `WAF.Profiles.Logging` 

**note:** if using the same datafeed for multiple WAF endpoints the log files will be stored together, they include a GUID in their file name so this should be safe

if using S3 you can use the following guide from AWS on setting up athena to query the logs
https://docs.aws.amazon.com/athena/latest/ug/waf-logs.html

## Motivation and Context

AWS WAF doesn't provide information out of the box about what caused a given request to fail. Instead they require you to setup a log forwarding solution to analyse the logs with an appropriate tool 

## How Has This Been Tested?
Tested locally across all components

## Types of changes

## Followup Actions
- Core - https://github.com/hamlet-io/engine/pull/1366
- AWS - https://github.com/hamlet-io/engine-plugin-aws/pull/105
- Bash Executor -  https://github.com/hamlet-io/executor-bash/pull/71

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/services/waf/resource.ftl
+++ b/aws/services/waf/resource.ftl
@@ -311,3 +311,113 @@
             }
     /]
 [/#macro]
+
+
+[#macro createWAFLoggingFromProfile occurrence wafaclId loggingProfile regional=false ]
+
+    [#local dataFeedArns = []]
+
+    [#if regional ]
+        [#local wafType = "regional" ]
+    [#else]
+        [#local wafType = "global" ]
+    [/#if]
+
+    [#list (loggingProfile.ForwardingRules)!{} as id,forwardingRule ]
+        [#list forwardingRule.Links?values as link]
+            [#if link?is_hash]
+                [#local linkTarget = getLinkTarget(occurrence, link) ]
+
+                [#if !linkTarget?has_content]
+                    [#continue]
+                [/#if]
+
+                [#local linkTargetRoles = linkTarget.State.Roles ]
+
+                [#if (linkTargetRoles.Outbound["logwatcher"]!{})?has_content
+                        && linkTarget.Direction == "outbound" ]
+
+                    [#switch linkTarget.Core.Type ]
+                        [#case DATAFEED_COMPONENT_TYPE ]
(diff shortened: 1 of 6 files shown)


aws/services/waf/resource.ftl
```diff
@@ -311,3 +311,113 @@
...
+                    [/#switch]
+                [/#if]
+            [/#if]
+        [/#list]
+    [/#list]
+
+    [@debug message="WAFLog" context=dataFeedArns enabled=true /]
+
+    [#if dataFeedArns?has_content ]
+        [#if deploymentSubsetRequired("cli", false) ]
```
ml019, line 367: Usually details of subsets are contained in calling setup routines and the service support routines assume they need to do whatever is provided. 

Maybe worth adding the actual subset names as parameters?

roleyfoley: So are you saying call this macro twice in each component? Once in the setup `cli` subset and once in the setup `epilogue`?

roleyfoley: So the other approach could be that we have `create` service routines and `setup` routines. Creates are the raw create an output things and `setup` is a more complete functionality which can include multiple subsets

ml019: The only change I was suggesting was to add the actual subset names as parameters. "cli" and "epilogue" are pretty well wired into the AWS support so I'll merge as is.

PR #106, opened 2020-08-07
chore(awstest): add tests for apigateway and s3

roleyfoley:
## Description
Add basic tests for apigateway and s3

## Motivation and Context
Supporting some basic tests and mock components for our standard serivces

## How Has This Been Tested?
Tested locally and used the tests 

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/awstest/scenarios/apigateway.ftl
+++ b/awstest/scenarios/apigateway.ftl
@@ -1,0 +1,126 @@
+[#ftl]
+
+[#macro awstest_scenario_apigateway ]
+
+    [@addDefinition
+        definition={
+            "appXapigatewaybase" : {
+                "swagger": "2.0",
+                "info": {
+                    "version": "v1.0.0",
+                    "title": "Proxy",
+                    "description": "Pass all requests through to the implementation."
+                },
+                "paths": {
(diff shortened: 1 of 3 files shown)


awstest/scenarios/apigateway.ftl
```diff
@@ -0,0 +1,127 @@
...
+                    "OutputSuffix" : "template.json",
+                    "Structural" : {
+                        "CFN" : {
+                            "Resource" : {
+                                "RestApi" : {
+                                    "Name" : "apiXappXapigatewaybase",
+                                    "Type" : "AWS::ApiGateway::RestApi"
+                                },
+                                "Deployment" : {
+                                    "Name" : "apiDeployXappXapigatewaybaseXj2xuero56t",
```
ml019, line 112: Are we fixing the runid? The deployment name id changes each time so feels like the test will fail...

roleyfoley: Good point, I've locked the runId in the https://github.com/hamlet-io/engine/pull/1373 and updated this PR to align with that fixed run Id

PR #115, opened 2020-08-18
Enable throttling apigatway at api, stage and method levels

RossMurr4y:
## Description

Implements hamlet-io/engine#1376 on the `apigateway` component. This is server-side throttling to protect the backend - it is not client-side throttling.

* Throttling defined at 3 independent levels: the whole API (aka defaults), the stage (overwriting any defaults) and individual methods (overwriting any of the others that would have applied to those methods)
* Throttling can be defined at two levels:
  * the stage
  * On a pattern, which implements the throttling on matching methods

Example `apigw.json` shows the following:

* stage level throttling is set to a burst of 5000, and a rate of 800
* Any methods patching the regex path `/.*/search` with a verb of `post` will be throttled to a burst of 1000 and rate of 300.
**Numbers are for example only, they are not appropriate throttle figures.**
```json
{
    "apigw" : {
        "Internal" : true,
        "Value" : {
            "Type" : "lambda",
            "Variable" : "IMP_API_LAMBDA",
            "BinaryTypes" : ["*/*"],
            "Throttling": {
                "BurstLimit" : 5000,
                "RateLimit" : 800
            },
            "Patterns" : [
                {
                    "Path" : "/.*/search",
                    "Verb" : "post",
                    "Variable" : "FLIBBERDY",
                    "BurstLimit" : 1000,
                    "RateLimit" : 300
                }
            ]
        }
    }
}
```

## Motivation and Context

Would like to be able to control the rate at which the API handles incoming requests to prevent overwhelming the backend in some scenarios.

## How Has This Been Tested?

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -305,26 +305,23 @@
         /]
 
         [#-- Throttling Configuration --]
-        [#local methodSettings = []]
         [#local openapiIntegrations = 
             getOccurrenceSettingValue(occurrence, [["apigw"], ["Integrations"]], true)]
 
-        [#if openapiIntegrations.Throttling?has_content]
-            [#local methodSettings = [
-                {
-                    "HttpMethod": "*",
-                    "ResourcePath": "/*",
-                    "LoggingLevel": "INFO",
-                    "DataTraceEnabled": true
-                } +
-                attributeIfContent(
-                    "ThrottlingBurstLimit", 
-                    openapiIntegrations.Throttling.BurstLimit) +
-                attributeIfContent(
-                    "ThrottlingRateLimit",
-                    openapiIntegrations.Throttling.RateLimit)
-            ]]
-        [/#if]
+        [#local methodSettings = [
+            {
+                "HttpMethod": "*",
+                "ResourcePath": "/*",
+                "LoggingLevel": "INFO",
+                "DataTraceEnabled": true
+            } +
+            attributeIfContent(
+                "ThrottlingBurstLimit", 
+                openapiIntegrations.Throttling.BurstLimit) +
+            attributeIfContent(
+                "ThrottlingRateLimit",
+                openapiIntegrations.Throttling.RateLimit)
+        ]]
 
         [#-- Integration Patterns (as Regex) into Matching Method Throttling (as explicit paths) --]
         [#if definitionsObject[core.Id]?? ]
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -324,7 +324,8 @@
         ]]
 
         [#-- Integration Patterns (as Regex) into Matching Method Throttling (as explicit paths) --]
-        [#if definitionsObject[core.Id]?? ]
+        [#if definitionsObject[core.Id]?has_content && 
+            openapiIntegrations.Throttling?has_content]
             [#list definitionsObject[core.Id].paths as path,pathConfig]
                 [#list pathConfig?keys as verb]
                     [#list openapiIntegrations.Patterns as pattern]
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -149,6 +149,7 @@
 
     [#local wafAclResources        = resources["wafacl"]!{} ]
     [#local cfResources            = resources["cf"]!{} ]
+    [#local planResources          = resources["plan"]!{} ]
     [#local customDomainResources  = resources["customDomains"]!{} ]
 
     [#local apiPolicyStatements    = _context.Policy ]
@@ -303,6 +304,13 @@
             outputs={}
             dependencies=apiId
         /]
+
+        [#-- Throttling Configuration --]
+        [#local throttleSettings = getApiThrottlingSettings(occurrence)]
+        [#local apiThrottle = throttleSettings.API]
+        [#local stageThrottle = throttleSettings.Versions[stageName]!{}]
+        [#local methodThrottles = throttleSettings.Methods?values]
+
         [@cfResource
             id=stageId
             type="AWS::ApiGateway::Stage"
@@ -316,7 +324,15 @@
                             "ResourcePath": "/*",
                             "LoggingLevel": "INFO",
                             "DataTraceEnabled": true
-                        }
+                        } +
+                        attributeIfContent(
+                            "ThrottlingBurstLimit", 
+                            stageThrottle,
+                            stageThrottle.BurstLimit!{}) +
+                        attributeIfContent(
+                            "ThrottlingRateLimit",
+                            stageThrottle,
+                            stageThrottle.RateLimit!{})
                     ],
                     "StageName" : stageName,
                     "AccessLogSetting" : {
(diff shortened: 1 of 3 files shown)


RossMurr4y:
Checks will of course fail until hamlet-io/engine#1376 is merged

aws/components/apigateway/state.ftl
```diff
@@ -83,6 +83,7 @@ created in either case.
 
     [#local certificatePresent = isPresent(solution.Certificate) ]
     [#local cfPresent          = isPresent(solution.CloudFront) ]
+    [#local cfPresent          = isPresent(solution.CloudFront) ]
```
ml019, line 86: Looks redundant?

aws/components/apigateway/setup.ftl
```diff
@@ -303,21 +303,57 @@
...
                     "RestApiId" : getReference(apiId),
-                    "MethodSettings": [
-                        {
-                            "HttpMethod": "*",
-                            "ResourcePath": "/*",
-                            "LoggingLevel": "INFO",
-                            "DataTraceEnabled": true
-                        }
-                    ],
+                    "MethodSettings": methodSettings,
```
ml019, line 356: We only want to set the defaults if there is throttling settings defined. so I think we want this to be added via `AttributeIfContent()`

aws/components/apigateway/setup.ftl
```diff
@@ -303,21 +303,57 @@
...
+
+        [#-- Throttling Configuration --]
+        [#local openapiIntegrations = 
+            getOccurrenceSettingValue(occurrence, [["apigw"], ["Integrations"]], true)]
+
+        [#local methodSettings = [
+            {
+                "HttpMethod": "*",
+                "ResourcePath": "/*",
+                "LoggingLevel": "INFO",
```
ml019, line 315: Don't want this I think unless explicitly defined - unless mandatory

aws/components/apigateway/setup.ftl
```diff
@@ -303,21 +303,57 @@
...
+                "DataTraceEnabled": true
```
ml019, line 316: Don't want this I think unless explicitly defined - unless mandatory

aws/components/apigateway/setup.ftl
```diff
@@ -303,21 +303,57 @@
...
+                openapiIntegrations.Throttling.RateLimit)
+        ]]
+
+        [#-- Integration Patterns (as Regex) into Matching Method Throttling (as explicit paths) --]
+        [#if definitionsObject[core.Id]?? ]
+            [#list definitionsObject[core.Id].paths as path,pathConfig]
+                [#list pathConfig?keys as verb]
+                    [#list openapiIntegrations.Patterns as pattern]
+                        [#if path?matches(pattern.Path)] 
+                            [#local methodSettings = combineEntities(
```
ml019, line 332: can just do array addition - I think it would be clearer.

ml019 (changes requested):

aws/components/apigateway/state.ftl
RossMurr4y: Yep, Not sure how that was added in there. Removed now.

aws/components/apigateway/setup.ftl
RossMurr4y: Yep certainly should be - fixed up

aws/components/apigateway/setup.ftl
RossMurr4y: Amended with a wrapper if-statement, and instantiating an empty array.

aws/components/apigateway/setup.ftl
RossMurr4y: Updated

aws/components/apigateway/setup.ftl
```diff
@@ -303,27 +303,64 @@
...
+                    openapiIntegrations.Throttling.RateLimit)
+            ]]
+        [/#if]
+
+        [#-- Integration Patterns (as Regex) into Matching Method Throttling (as explicit paths) --]
+        [#if definitionsObject[core.Id]?? ]
+            [#list definitionsObject[core.Id].paths as path,pathConfig]
+                [#list pathConfig?keys as verb]
+                    [#list openapiIntegrations.Patterns as pattern]
+                        [#if path?matches(pattern.Path)]
```
ml019, line 334: Needs verb comparison as well

aws/components/apigateway/setup.ftl
```diff
@@ -303,27 +303,64 @@
...
+            ]]
+        [/#if]
+
+        [#-- Integration Patterns (as Regex) into Matching Method Throttling (as explicit paths) --]
+        [#if definitionsObject[core.Id]?? ]
+            [#list definitionsObject[core.Id].paths as path,pathConfig]
+                [#list pathConfig?keys as verb]
+                    [#list openapiIntegrations.Patterns as pattern]
+                        [#if path?matches(pattern.Path)] 
+                            [#local methodSettings +=
```
ml019, line 335: Should only generate entry if the throttling element is defined in the pattern.

aws/components/apigateway/setup.ftl
```diff
@@ -303,27 +303,64 @@
             outputs={}
             dependencies=apiId
         /]
+
+        [#-- Throttling Configuration --]
+        [#local methodSettings = []]
+        [#local openapiIntegrations = 
+            getOccurrenceSettingValue(occurrence, [["apigw"], ["Integrations"]], true)]
+
+        [#if openapiIntegrations.Throttling?has_content]
```
ml019, line 312: My bad. Looking at original code, we are always creating at least one methodSetting entry so this code doesn't need to be wrapped in the check for throttling.

aws/components/apigateway/setup.ftl
```diff
@@ -303,27 +303,64 @@
...
-                            "DataTraceEnabled": true
-                        }
-                    ],
                     "StageName" : stageName,
                     "AccessLogSetting" : {
                         "DestinationArn" : getArn(accessLgId),
                         "Format" : "$context.identity.sourceIp $context.identity.caller $context.identity.user $context.identity.userArn [$context.requestTime] $context.apiId $context.httpMethod $context.resourcePath $context.protocol $context.status $context.responseLength $context.requestId"
                     }
                 } +
+                attributeIfContent("MethodSettings", methodSettings) +
```
ml019, line 363: Again, my bad. We always want this attribute but I think we leave the code as you now have it. It works the same and would permit a case where we didn't want to always provide method defaults.

ml019 (changes requested):

aws/components/apigateway/setup.ftl
RossMurr4y: Updated: 15552a5

aws/components/apigateway/setup.ftl
RossMurr4y: Updated: 8f2a517

aws/components/apigateway/setup.ftl
RossMurr4y: Updated: 464cde1

PR #140, opened 2020-10-21
Feat apigw kinesis logging

RossMurr4y:
## Description

* Enable APIGateway component to log access logs to the OpsData bucket with a Kinesis Firehose
* Implemented Kinesis Data Stream resources (unused, as weren't necessary in the end)
* Refactored WAF to use the same logging functionality as the API Gateway
   * This involved also updating WAF for `cdn` and `lb` components

## Motivation and Context

Would like to be able to process a large amount of invocation logs with a Kinesis Firehose. This is a part of a greater push to revisit the way logging works in general.

## How Has This Been Tested?

Locally tested and deployed successfully.

## Types of changes

## Followup Actions

- [ ] Requires https://github.com/hamlet-io/engine/pull/1417

## Checklist:

- [x] None of the above.


aws/components/lambda/state.ftl
```diff
@@ -128,7 +128,8 @@
                     "default" : "invoke",
                     "invoke" : lambdaInvokePermission(id),
                     "authorise" : lambdaInvokePermission(id),
-                    "authorize" : lambdaInvokePermission(id)
+                    "authorize" : lambdaInvokePermission(id),
+                    "kinesis" : lambdaKinesisPermission(id)
```
ml019, line 132: Can we come up with a role which is less tied to the resource type e.g. "stream"?

ml019 (changes requested):
With merging of support for service to resource mapping, there will need to be a bit of extra config added for AWS_KINESIS_DATA_STREAM_RESOURCE_TYPE

aws/components/s3/setup.ftl
```diff
@@ -203,21 +202,11 @@
                     [/#if]
                     [#break]
 
+                [#case "baselinedata"]
```
roleyfoley, line 205: Better to use the component type Variable instead of fixed value. Shows the meaning of what this is a bit more

aws/services/kinesis/resource.ftl
```diff
@@ -45,6 +84,149 @@
...
+                [#break]
+
+        [/#switch]
+
+        [#-- [@createKinesisDataStream 
+            id=dataStream.Id
+            name=dataStream.Name
+            keyId=dataStreamEncryptionKey
+            dependencies=dependencies
+        /] --]
```
roleyfoley, line 208: I don't think we should include kinesis streams as part of this process. This macro is to create a firehose not a data stream

aws/components/lambda/state.ftl
roleyfoley: Also need to make the distinction between the different kiniesis streams if we are adding support for datastreams. Whats the purpose of this role anyway?

roleyfoley (changes requested):

RossMurr4y:
@ml019 @roleyfoley I've addressed your comments in 37d803b, 6c9ce96 & 3e4b92d.

I found the AWS documentation on Data and Delivery streams quite confusing and contradictory. 

- In the Kinesis Firehose documentation, the Delivery Streams discuss receiving data either by Direct PUT requests, or by Data Stream. There are also some places that used "Firehose Data Stream" and "Delivery Stream" interchangeably. This led me to believe that the Kinesis Data Stream was necessary, so I implemented it.
- In the API Gateway documentation, there are tutorials that have the APIGW communicate directly with the Firehose Delivery Stream. I couldn't get my tests to work however, but I could get the "test data" from the Firehose to the intended bucket.
- Eventually I worked out that the problem was that using the "test" calls within the API Gateway do not produce logs, so it was my testing and not implementation that was at fault.
- When I reversed some of what I'd implemented (dropping their git commits) it looks like some of that was caught up in other commits so was left behind (like the commented-out datastream.

So, I've left the macros for the creation of a Data Stream implemented should we want it in the future, but I've removed the remainder of my older implementations (including the lambda role).

RossMurr4y:
This has been updated to include refactorings for WAF logging, so that it uses the same implementation of implementing the Kinesis Firehose across components.

aws/components/apigateway/id.ftl
```diff
@@ -1,7 +1,18 @@
...
 [@addResourceGroupInformation
     type=APIGATEWAY_COMPONENT_TYPE
-    attributes=[]
+    attributes=[
+        {
+            "Names" : "AccessLogging",
+            "Children" : [
+                {
+                    "Names" : "DestinationLink",
+                    "Description" : "Destination for the Execution logs. If not provided, Execution logs will be stored in the OpsData DataBucket.",
```
roleyfoley, line 10:
```suggestion
                    "Description" : "Destination for the Access logs. If not provided, Execution logs will be stored in the OpsData DataBucket.",
```

aws/components/cdn/setup.ftl
```diff
@@ -478,11 +478,23 @@
...
+            errorPrefix=formatRelativePath(occurrence.Core.FullRelativePath, "waf", "error")
+            streamNamePrefix="aws-waf-logs-"
+        /]
+
+        [@createWAFLoggingDeliveryStream
             wafaclId=wafAclId
-            loggingProfile=wafLoggingProfile
+            deliveryStreamId=wafFirehoseStreamId
             regional=false
         /]
```
roleyfoley, line 500: Should this check for `[#if solution.WAF.Logging.Enabled`

aws/components/lb/setup.ftl
```diff
@@ -722,6 +722,25 @@
...
+                    bucketPrefix=formatRelativePath(occurrence.Core.FullRelativePath, "waf")
+                    errorPrefix=formatRelativePath(occurrence.Core.FullRelativePath, "waf", "error")
+                    streamNamePrefix="aws-waf-logs-"
+                /]
+
+                [@createWAFLoggingDeliveryStream
+                    wafaclId=wafAclResources.acl.Id
+                    deliveryStreamId=wafFirehoseStreamId
+                    regional=false
+                /]
```
roleyfoley, line 742: Should this check for `[#if solution.WAF.Logging.Enabled`

roleyfoley (changes requested):

aws/components/lb/setup.ftl
RossMurr4y: Corrected in 03237109, cheers

aws/components/cdn/setup.ftl
RossMurr4y: Yep thanks, fixed up in 1cf5d5be

aws/components/apigateway/id.ftl
RossMurr4y: Thanks, have amended db4ed004

aws/components/s3/setup.ftl
```diff
@@ -202,6 +202,7 @@
                     [/#if]
                     [#break]
 
+                [#case BASELINE_DATA_COMPONENT_TYPE ]
```
ml019, line 205:
```suggestion
```

aws/components/apigateway/setup.ftl
```diff
@@ -252,15 +260,37 @@
             ] ]
     [/#if]
 
-    [#local accessLgId   = resources["accesslg"].Id]
-    [#local accessLgName = resources["accesslg"].Name]
+    [#if solution.AccessLogging.Enabled]
+        [#-- Manage Access Logs with Kinesis Firehose --]
```
ml019, line 264: Need an extra attribute to control whether firehose or cloud watch log groups is used. Also need logic to permit existing installations to preserve the existing cloudwatch log groups.

So I think something like `"aws:Firehose: True"` and `"aws:KeepLogGroup" : true`. stageLogTarget is controlled via the former. The creation of the log group is performed if either Firehose=false or KeepLogGroup=true.

Then we can incrementally remove the log groups once the data has been processed via firehose as a separate activity.

aws/components/apigateway/id.ftl
```diff
@@ -1,7 +1,18 @@
 [#ftl]
 [@addResourceGroupInformation
     type=APIGATEWAY_COMPONENT_TYPE
-    attributes=[]
+    attributes=[
+        {
+            "Names" : "AccessLogging",
+            "Children" : [
```
ml019, line 7: See below re adding a couple of extra children to switch between firehose and log groups.

ml019 (changes requested):

aws/components/apigateway/id.ftl
RossMurr4y: @ml019 I have incorporated your changes. Could you take another look?

roleyfoley (approved):
All requested fixes added

PR #142, opened 2020-10-21
feat(awsdiagrams): adds diagram mappings for aws resources

roleyfoley:
## Description
Implements the required resource and service mapping for https://github.com/hamlet-io/engine-plugin-diagrams/pull/7 in the diagrams provider 

This has been implemented as a separate provider under the aws plugin, this ensures that the provider is maintained in line with changes to the resources and services available within our aws provider. Note that this provider depends on the aws provider and should be loaded with the aws provider when its is required

## Motivation and Context
When drawing diagrams being able to show the icons for a particular provider as specifically as possible allows you to provide more detailed diagrams using pictures to represent as much as possible. 

In order to do this we need to map our resource type definitions to the classes available for our diagrams provider 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] Depends on https://github.com/hamlet-io/engine-plugin-diagrams/pull/7 

## Checklist:
- [X] None of the above.


ml019 (approved):
Nicely extensible but doesn't get in the way if you don't want diagrams.

PR #144, opened 2020-10-22
fix(networkacl): use the id instead of existing ref for lookups

roleyfoley:
## Description
Changes the reference value provided for security groups provided the networkacl role. The security group will now be provided as the resource Id instead of the evaluated output provided by getExistingReference

## Motivation and Context
When linking between two components in the same deployment unit the references returned were of the groupId for an existing deployment. As a result of this the deployment would fail if the group ids changed during a deployment or wouldn't return correctly in a new deployment

## How Has This Been Tested?
Tested locally on a development deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/bastion/state.ftl
+++ b/aws/components/bastion/state.ftl
@@ -52,14 +52,14 @@
             "Roles" : {
                 "Inbound" : {
                     "networkacl" : {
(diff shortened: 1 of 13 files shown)


ml019 (approved):
No suggestions for improvement

PR #147, opened 2020-10-28
WIP - Feature - Baseline DataBucket Replication

RossMurr4y:
## Description

Adds the same component roles and the permissions/policies that exist on the `s3` component to the `baselinedata` subcomponent type.

Note: a lookup of the baseline `key` subcomponent to obtain the cmk is currently performed by the `s3` component so that it can receive appropriate permissions on another bucket. This behaviour cannot directly be copied to the `baselinedata` component because doing so will cause the links between components to infinitely look up eachother. Because of this I have extracted the logic into its own function, and the `s3` component when linking to a databucket with those roles will evaluate the function during setup.

## Motivation and Context

The `baselinedata` subcomponent type is a specific implementation of the `s3` component. It should grant the same roles and permissions from a link that would occurr if linking to an `s3` component.

## How Has This Been Tested?

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [ ] None of the above.


roleyfoley:
@RossMurr4y is this PR still required with #154 now raised as well?

PR #150, opened 2020-10-29
feat: add compute provider support to ecs host

roleyfoley:
## Description

Adds initial support for compute providers as part of the ecs host in line with https://github.com/hamlet-io/engine/pull/1442 

This initial support only sets the default capacity provider for an ECS host which is used when a capacity provider or launch type have not been explicitly set, which should maintain compatibility with our existing deployments 

By default all ECS clusters will have 3 capacity providers available, FARGATE, FARGATE_SPOT and one based on the EC2 Autoscale group which is provisioned as part of the component. 

The default ComputeProvider uses the EC2 Autoscale group for container placements, this is inline with the standard behaviour we have now. Some extra ComputeProvider profiles have been included ( fargate, fargate-add-spot, fargatespot ) have been added to show different strategies that you might want to use 

Due to some circular dependency issues two template passes are required to complete the full setup. There is an issue where you need to pass the ECS Cluster name to the ASG for registration, but the ECS Cluster needs to know the Capacity Provider details, which need to know the ASG Name. This is being tracked in https://github.com/aws/containers-roadmap/issues/631 which has some details on the issue

## Motivation and Context
The main motivation to implement this in the AWS ecs provider is to be able to take advantage of the FARGATE_SPOT offering which allows for significant cost savings on services which can handle interruptions 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] Depends on https://github.com/hamlet-io/engine/pull/1442

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/ecs/setup.ftl
+++ b/aws/components/ecs/setup.ftl
@@ -1,6 +1,6 @@
 [#ftl]
 [#macro aws_ecs_cf_deployment_generationcontract_solution occurrence ]
-    [@addDefaultGenerationContract subsets=[ "template", "cli", "epilogue" ] /]
+    [@addDefaultGenerationContract subsets=[ "template" ] /]
 [/#macro]
 
 [#macro aws_ecs_cf_deployment_solution occurrence ]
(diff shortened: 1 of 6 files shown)


roleyfoley:
converting to draft to align with https://github.com/hamlet-io/engine/issues/1444

ml019 (approved):
Nice to see a lot of functionality more into the template. The idea of a default and additional capacity provides a logical separation for different strategies.

roleyfoley:
Will need to merge this now that the shared provider has been merged in

PR #155, opened 2020-11-03
Fix state routine links

RossMurr4y:
## Description

Updates all state routine calls to `getBaselineLinks` function to allow for the return of an empty result.

## Motivation and Context

Due to recent changes the current implementation will not allow you to include one of these components using this function in the state routine if the baseline deployment unit is not yet deployed.

## How Has This Been Tested?

Local template generation.

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


aws/components/objectsql/state.ftl
```diff
@@ -32,7 +32,7 @@
                     "consume" :
                         athenaConsumePermission(workGroupId) +
                         s3AllPermission(baselineComponentIds["AppData"], getAppDataFilePrefix(occurrence))
```
roleyfoley, line 34:
```suggestion
                        s3AllPermission(dataBucket, getAppDataFilePrefix(occurrence))
```
Looks like that variable isn't being used

roleyfoley (changes requested):

roleyfoley (approved):

PR #156, opened 2020-11-04
feature - LB Error on Missing Cert

RossMurr4y:
## Description

Adds a fatal error message if a certificate is defined for an ALB but the reference could not be found.

## Motivation and Context

Currently it will return an empty string and look valid until stack deployment fails. This provides the failure warning earlier allowing a user to correct it.

## How Has This Been Tested?

Local template generation.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


aws/services/lb/resource.ftl
```diff
@@ -203,6 +203,18 @@
         [#local protocol = port.Protocol]
     [/#if]
 
+    [#local certificateArn = getReference(certificateId, ARN_ATTRIBUTE_TYPE, regionId)]
```
roleyfoley, line 206: Should probably make this getExistingReference instead of getReference, otherwise it could return the ref object I think

roleyfoley (changes requested):

aws/services/lb/resource.ftl
RossMurr4y: Cheers, sorted in 32b59f9

aws/services/lb/resource.ftl
```diff
@@ -222,12 +234,12 @@
...
                     "Certificates" : [
                         {
-                            "CertificateArn" : getReference(certificateId, ARN_ATTRIBUTE_TYPE, regionId)
+                            "CertificateArn" : certificateArn
                         }
                     ],
                     "SslPolicy" : sslPolicy
                 },
-                port.Certificate!false)
+                (port.Certificate!false && certificateArn?has_content))
```
roleyfoley, line 242: I don't think we need to exclude including the resource content if we are going to throw a fatal error. Debugging issues would be a bit weird if that was the case

roleyfoley (changes requested):

aws/services/lb/resource.ftl
RossMurr4y: Yep, i was 50/50 on that one. I'll remove it.

RossMurr4y: Removed in efe8bb8

PR #157, opened 2020-11-04
Feat datafeed to Baseline Data

RossMurr4y:
## Description

Enables configuring a `datafeed` component with a baseline data bucket as the destination link.

## Motivation and Context

Would like to be able to log directly to the `opsdata` baseline data bucket with a `datafeed`.

## How Has This Been Tested?

Local template generation.

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.


roleyfoley:
Whats the user case for this? the opsdata bucket has very specific prefixes which are given to each of the components. So if you were to do this you  would need to set a fixed prefix rather than allow the data feed to define it.

RossMurr4y:
> Whats the use case for this? the opsdata bucket has very specific prefixes which are given to each of the components to store their own data.
> 
> So if you were to do this you would need to set a fixed prefix rather than allow the data feed to define it.

The LogForwarding Module was the use case here. I queried @ml019 about the naming for the bucket the module would create, and the discussion then led us to question if the module should implement any `s3` component or instead it's `datafeed` should log to the local segment `opsdata` and from there we would migrate files into an a Tenant level `s3` bucket.

Happy to go either way, or even to just have the `datafeed` link direct to the tenant-level bucket.

roleyfoley:
The ops data bucket stores more than logs, it also stores asFile settings like certificates and other app specific data files so I'm not sure how we are going to go just using that bucket 

The other point here is that layout of the opsdata bucket is fixed based on the pattern 

`<data type>/<component FullAbsolutePath>` 

E.g. an opsdata bucket for a product with load balancers and as file data has the following root level folders 
Screen Shot 2020-11-05 at 10 40 31

Where the AWSLogs prefix is locked by AWS and is used to store the load balancer logs 
The settings we store as files is under `settings` 

So I think we need to say that if the link is to an opsdata bucket the Prefix datafeed prefix is prepended to the absoluteFull Path of the data feed component

RossMurr4y:
> The ops data bucket stores more than logs, it also stores asFile settings like certificates and other app specific data files so I'm not sure how we are going to go just using that bucket
> 
> The other point here is that layout of the opsdata bucket is fixed based on the pattern
> 
> `<data type>/<component FullAbsolutePath>`
> 
> E.g. an opsdata bucket for a product with load balancers and as file data has the following root level folders
> Screen Shot 2020-11-05 at 10 40 31
> 
> Where the AWSLogs prefix is locked by AWS and is used to store the load balancer logs
> The settings we store as files is under `settings`
> 
> So I think we need to say that if the link is to an opsdata bucket the Prefix datafeed prefix is prepended to the absoluteFull Path of the data feed component

I believe I've addressed this now in 0bd0c4d - @roleyfoley would you mind taking a look?

aws/components/datafeed/setup.ftl
```diff
@@ -238,15 +238,23 @@
...
 
+        [#-- Establish bucket prefixes --]
+        [#local streamS3DestinationPrefix = solution.Bucket.Prefix]
+        [#local streamS3DestinationErrorPrefix = solution.Bucket.ErrorPrefix]
 
         [#switch (destinationLink.Core.Type)!"notfound" ]
             [#case BASELINE_DATA_COMPONENT_TYPE]
+                [#if destinationLink.Core.SubComponent.Name == "opsdata"]
+                    [#local streamS3DestinationPrefix = formatPath(true, core.FullAbsolutePath, solution.Bucket.Prefix)]
+                    [#local streamS3DestinationErrorPrefix = formatPath(true, core.FullAbsolutePath, solution.Bucket.ErrorPrefix)]
```
roleyfoley, line 249: Other way around 

So the prefix defined on the Datafeed would define what type of data it is, CWLogs for example

roleyfoley (changes requested):

aws/components/datafeed/setup.ftl
```diff
@@ -238,14 +238,23 @@
                                                 cmkKeyId,
                                                 streamBackupLoggingConfiguration )]
 
+        [#-- Establish bucket prefixes --]
+        [#local streamS3DestinationPrefix = solution.Bucket.Prefix]
+        [#local streamS3DestinationErrorPrefix = solution.Bucket.ErrorPrefix]
```
roleyfoley, line 243: Should also use these at 248/249 if the intention was to show sharing

roleyfoley (changes requested):

aws/components/datafeed/setup.ftl
RossMurr4y: Corrected in ff090c0

aws/components/datafeed/setup.ftl
RossMurr4y: Gotcha, fixed up in ff090c0

aws/components/datafeed/setup.ftl
```diff
@@ -238,14 +238,23 @@
...
 
         [#switch (destinationLink.Core.Type)!"notfound" ]
+            [#case BASELINE_DATA_COMPONENT_TYPE]
+                [#if destinationLink.Core.SubComponent.Name == "opsdata"]
+                    [#local streamS3DestinationPrefix = formatPath(true, streamS3DestinationPrefix, core.FullAbsolutePath)]
```
roleyfoley, line 248:
```suggestion
                    [#local streamS3DestinationPrefix = formatRelativePath(streamS3DestinationPrefix, core.FullAbsolutePath)]
```

S3 doesn't support Absolute paths, only Relative paths 
In S3 world `/` doesn't actually mean folder, they just use them to create folder like display in the console

aws/components/datafeed/setup.ftl
```diff
@@ -238,14 +238,23 @@
...
+                    [#local streamS3DestinationErrorPrefix = formatPath(true, streamS3DestinationErrorPrefix, core.FullAbsolutePath)]
```
roleyfoley, line 249:
```suggestion
                    [#local streamS3DestinationErrorPrefix = formatRelative(streamS3DestinationErrorPrefix, core.FullAbsolutePath)]
```

S3 doesn't support Absolute paths, only Relative paths
In S3 world / doesn't actually mean folder, they just use them to create folder like display in the console

roleyfoley (changes requested):

RossMurr4y:
Thanks @roleyfoley for your commit suggestions - I've committed  them now.

roleyfoley (approved):

PR #158, opened 2020-11-05
fix(logstreaming): fixes to logstreaming setup

roleyfoley:
## Description
A collection of changes to the log streaming setup for access logs and WAF logs 
- Moves the resource definitions into the state definition to allow for them to be discovered by tools which use component states 
- Updates the naming of the firehose stream to ensure they are unique - previously this was based on the link destination Id which causes problems when we have multiple components forwarding to the same destination, such as the ops data link 
- Truncates stream naming to 64 chars to align with the requirements of the resource naming
- Add subset checks during the setup macro to ensure that we align with resource sets 
- Defines fixed prefixes for opsdata forwarding to avoid conflicts on data type and logging component 
- Updates the WAF logging setup to pass the Id to the script function and replace the existing cli based configuration requirements 

## Motivation and Context
These issues were found on testing an existing deployment which had WAF logging enabled 

## How Has This Been Tested?

## Types of changes

## Followup Actions
- [x] Requires: https://github.com/hamlet-io/executor-bash/pull/125 

## Checklist:
- [X] None of the above.


ml019 (approved):
No issues. Expect further testing in client dpeloyments.

PR #165, opened 2020-11-11
New Module - Log Consolidator

RossMurr4y:
## Description

This PR implements solution-wide log consolidation into a standardised format which is then ready for ingestion into ElasticSearch.

Components with standard logging behaviour to CloudWatch are monitored by a “LogWatcher” data feed using Kinesis Firehose. The logs are then processed by a Lambda Function to filter and process the logs into destination locations based on the log type. Once processed and sorted by type, the logs can be ingested by 3rd party tooling such as ElasticSearch for solution-wide log analysis.

Components that have non-standard logging behaviour such as the API Gateway receive their own Kinesis Firehose stream, which then outputs its logs to a fixed destination. From there the logs are replicated into the consolidation log store.

Unless provided new parameters for a log processing lambda, the default implementation will retrieve and use the [hamlet lambda-log-processor](https://github.com/hamlet-io/lambda-log-processors/releases/download/v1.0.2/cloudwatch-firehose.zip)

## Motivation and Context

It is desirable to be able to have all logs across an entire layer consolidated into one location. These can then be fed into a tool such as ElasticSearch or a SIEM.

## How Has This Been Tested?

Tested at the product and tenant levels.

## Types of changes

## Followup Actions

- [ ]  Add test cases as per the to-dos

## Checklist:

- [x] None of the above.


aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                                        "Logging" : "consolidate"
+                                    }
+                                }
+                            },
+                            "cdn" : {
+                                "EnableLogging" : true,
+                                "WAF" : {
+                                    "Profiles" : {
+                                        "EnableLogging" : true,
+                                        "Logging" : "consolidate"
```
ml019, line 204: I thought WAF didn't need logging profiles any more?

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                                    }
+                                }
+                            },
+                            "cdn" : {
+                                "EnableLogging" : true,
+                                "WAF" : {
+                                    "Profiles" : {
+                                        "EnableLogging" : true,
```
ml019, line 203: This doesn't look like a profile?

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                    "Components" : {
+                        datafeedName : {
+                            "datafeed": {
+                                "Instances": {
+                                    "default": {
+                                        "DeploymentUnits": [ datafeedName ]
+                                    }
+                                },
+                                "Encrypted": true,
+                                "Destination": {
```
ml019, line 82: How is this different to the "store" link below?

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                                    "Link": {
+                                        "Tier": "mgmt",
+                                        "Component": "baseline",
+                                        "DataBucket": "opsdata",
+                                        "Instance": "",
+                                        "Version": ""
+                                    }
```
ml019, line 89: IS there any path control as to what top level prefix the logs will end up under?

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Yep you're right - I was going of the Schema that is on the hamlet.io site which needs to be setup with CI to update on its own. I should have known this though, given I removed them! 

I'll fix this one up now.

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Resolved by the removal of WAF profiles

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+        {
+            "Names" : "namePrefix",
+            "Type" : STRING_TYPE,
+            "Description" : "A prefix appended to component names and deployment units to ensure uniquness",
+            "Default" : "log"
+        },
+        {
+            "Names" : "deploymentProfile",
+            "Type" : STRING_TYPE,
+            "Description" : "The deployment profile id to enable log consolidation on."
```
roleyfoley, line 17: Should either add a `Default` or `Manadatory` flag her

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+        settingSets=[
+           {
+                "Type" : "Settings",
+                "Scope" : "Products",
+                "Namespace" : lambdaSettingNamespace,
+                "Settings" : {
+
+                }
+            } 
+        ]
```
roleyfoley, line 69: Shouldn't be required if you aren't injecting any settings

aws/modules/consolidate-logs/module.ftl
roleyfoley: The store link shouldn't be required, the destination link should be able to do setup the appropriate permissions to get this going

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                                        "Handler": "src/run.lambda_handler",
+                                        "Links": {
+                                            "store": {
+                                                "Tier": "mgmt",
+                                                "Component": "baseline",
+                                                "DataBucket" : "opsdata",
+                                                "Instance" : "",
+                                                "Version": "",
+                                                "Role": "datafeed"
+                                            }
```
roleyfoley, line 144: I don't think the lambda function has any knowlege of the S3 bucket only the firehose should have that data

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                            "Filter": "all-logs",
+                            "Links": {
+                                "store": {
+                                    "Tier": "mgmt",
+                                    "Component": "baseline",
+                                    "Instance": "",
+                                    "Version": "",
+                                    "DataBucket" : "opsdata",
+                                    "Role" : "produce"
+                                }
```
roleyfoley, line 171: I think that link should be to the datafeed rather than the opsdata bucket

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+                                        "Instance": "",
+                                        "Version": ""
+                                    }
+                                },
+                                "LogWatchers": {
+                                    "app-all": {
+                                        "LogFilter": "all-logs"
+                                    }
+                                },
+                                "aws:WAFLogFeed": true,
```
roleyfoley, line 96: If this is the cwlog collection datafeed then it shouldn't have WAF prefix applied to it

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,226 @@
...
+    namePrefix
+    deploymentProfile
+    lambdaSourceUrl
+    lambdaSourceHash
+    tier]
+
+    [@debug message="Entering Module: consolidate-logs" context=layerActiveData enabled=false /]
+
+    [#local lambdaName = formatName(namePrefix, "lambda")]
+    [#local datafeedName = formatName(namePrefix, "datafeed")]
```
roleyfoley, line 50: Should we maybe add extra prefix here or instances on the the component for cwlogs ( this datafeed + processor) will only support logs sent to it from cloudwatch logs

roleyfoley (changes requested):

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Whilst there is separate link and destinationlink processing, there is no reason to do both for the target as I had first thought. I've removed it now

aws/modules/consolidate-logs/module.ftl
roleyfoley: The datafeed component has a configuration option to control the two prefixes required for datafeed 

 Bucket.Prefix - Where the actual data will land 
Bucket.ErrorPrefix - Where failed data will land ( this is usually used with non S3 backends but still needs to be configured ) 

Should we make this Prefix `CWLogs/Logs/`  and ErrorPrefix `CWLogs/Errors/`  ? with https://github.com/hamlet-io/engine-plugin-aws/pull/157/files that will also include the data feed full path after the initial prefix so we can have multiple cwlogdata feeds if required

roleyfoley (changes requested):

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Yep cheers - have tidied up.
I had anticipated needing some for the lambda but they weren't necessary.

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Cheers, removed.

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Yep, frustratingly I had this right the first time around and I replicated it poorly into the module

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Updated

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Good idea, makes it clearer. Have added that in now.

aws/modules/consolidate-logs/module.ftl
RossMurr4y: I've gone with `Mandatory` as it feels safer to have the user define what they want than have them only half-evaluate if any given default group is appropriate

aws/modules/consolidate-logs/module.ftl
RossMurr4y: I have added this in https://github.com/hamlet-io/engine-plugin-aws/pull/165/commits/79c8e98b10b059fed45b7b9bb78c9e756229eff4 - we can discuss further once we see how this handles.

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,210 @@
...
+
+    [#local lambdaName = formatName(namePrefix + "cwlogslambda")]
+    [#local datafeedName = formatName(namePrefix + "cwlogsdatafeed")]
+
+    [#local product = getActiveLayer(PRODUCT_LAYER_TYPE) ]
+    [#local environment = getActiveLayer(ENVIRONMENT_LAYER_TYPE)]
+    [#local segment = getActiveLayer(SEGMENT_LAYER_TYPE)]
+
+    [#local namespace = formatName(product["Name"], environment["Name"], segment["Name"])]
+    [#local lambdaSettingNamespace = formatName(namespace, lambdaName)]
```
roleyfoley, line 58: I don't think you need these if you've removed the settings injection

roleyfoley (changes requested):

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Cheers, tidied them up in 8aed1d6

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,197 @@
...
+                                    "aws:KeepLogGroup" : true
+                                },
+                                "WAF" : {
+                                    "Logging" : {
+                                        "Enabled" : true
+                                    }
+                                }
+                            },
+                            "cdn" : {
+                                "EnableLogging" : true
```
ml019, line 177: Do we still need the WAF Logging true on the cdn - it may be the default but worth being explicit

ml019 (changes requested):

aws/modules/consolidate-logs/module.ftl
RossMurr4y: Fixed up

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,202 @@
...
+                            "cdn" : {
+                                "EnableLogging" : true,
+                                "WAF" : {
+                                    "Logging" : {
+                                        "Enabled" : true
+                                    }
+                                }
+                            },
+                            "lb" : {
+                                "Logs" : true
```
roleyfoley, line 185: one last bit enable WAF where WAF is 
```suggestion
                                "Logs" : true
                                "WAF" : { 
                                    "Logging" : { 
                                         "Enabled" : true
                                     }
                                 }
```

roleyfoley (changes requested):

aws/modules/consolidate-logs/module.ftl
```diff
@@ -0,0 +1,207 @@
...
```
roleyfoley, line 185:
```suggestion
                                "Logs" : true,
```

roleyfoley (dismissed):

roleyfoley (approved):

PR #166, opened 2020-11-11
feat: private bastion

ml019:
## Description
Create a bastion that is only accessible via SSM.

## Motivation and Context
If no possibility of access from outside the vpc is desired, this module disables the bastion that is included in the masterdata and replaces it with a private bastion. The tier specified needs to be private.

## How Has This Been Tested?
Local deployment

## Types of changes

## Followup Actions
- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/modules/privatebastion/module.ftl
+++ b/aws/modules/privatebastion/module.ftl
@@ -1,0 +1,91 @@
+[#ftl]
+
+[@addModule
+    name="privatebastion"
+    description="Bastion access via SSM only"
+    provider=AWS_PROVIDER
+    properties=[
+        {
+            "Names" : "tier",
+            "Type" : STRING_TYPE,
+            "Description" : "The tier to use to host the private bastion",
+            "Default" : "msg"
+        },
+        {
+            "Names" : "component",
+            "Type" : STRING_TYPE,
+            "Description" : "The component to use to host the private bastion",
+            "Default" : "ssh"
+        },
+        {
+            "Names" : "deploymentUnit",
+            "Type" : STRING_TYPE,
+            "Description" : "The deployment unit for the private bastion",
+            "Default" : "ssh"
+        },
+        {
+            "Names" : "multiAZ",
+            "Type" : BOOLEAN_TYPE,
+            "Description" : "Multi-AZ support on the private bastion",
+            "Default" : true
+        }
+    ]
+/]
+
+[#macro aws_module_privatebastion
+            tier
+            component
+            deploymentUnit
+            multiAZ ]
+
+    [@loadModule
+        blueprint={
+            "Tiers" : {
+                tier : {
+                    "Components" : {
+                        component: {
+                            "DeploymentUnits": [
+                                deploymentUnit
+                            ],
+                            "MultiAZ": multiAZ,
+                            "bastion": {
+                                "AutoScaling": {
+                                    "DetailedMetrics": false,
+                                    "ActivityCooldown": 180,
+                                    "MinUpdateInstances": 0,
+                                    "AlwaysReplaceOnUpdate": false
+                                },
+                                "Permissions": {
+                                    "Decrypt": true
+                                }
+                            }
+                        }
+                    }
+                },
+                "mgmt" : {
+                    "Components" : {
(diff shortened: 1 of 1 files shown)


aws/modules/privatebastion/module.ftl
```diff
@@ -0,0 +1,91 @@
...
+[@addModule
+    name="privatebastion"
+    description="Bastion access via SSM only"
+    provider=AWS_PROVIDER
+    properties=[
+        {
+            "Names" : "tier",
+            "Type" : STRING_TYPE,
+            "Description" : "The tier to use to host the private bastion",
+            "Default" : "msg"
```
roleyfoley, line 12: Not sure the msg tier is the best default tier for this

roleyfoley (changes requested):

roleyfoley:
Should we actually just make this the bastion component with a default on the mgmt tier, then remove the bastion component from the master data and then load this module in the master data

ml019:
Converting the current bastion to a module seems a logical step - occurred to me as I was doing it.

But loading in master data still means you can't get rid of it ...?

aws/modules/privatebastion/module.ftl
ml019: I didn't want to introduce another tier just for this. Open to alternatives from the existing set ...

roleyfoley:
Yeah you can disable a module even when it's specified in the modules list

ml019:
Interesting, so we could evolve the current master data into a few modules that get loaded by default as a transition to removal of them from the masterdata altogether

aws/modules/privatebastion/module.ftl
RossMurr4y: Any that are in the existing set will already have the possibility of something already being deployed to them - by their nature they are common enough to be pre-defined. Providing one over another is going to enable some projects to deploy without disruption, whilst others will still need to. 

If you define your own by default, you can at least make it clear that this will create a new tier and subsequent subnet unless an override is provided. Otherwise its murky, and depends on a number of solution-specific factors as to the best tier.

aws/modules/privatebastion/module.ftl
```diff
@@ -0,0 +1,91 @@
...
+        }
+    ]
+/]
+
+[#macro aws_module_privatebastion
+            tier
+            component
+            deploymentUnit
+            multiAZ ]
+
```
RossMurr4y, line 40: If the tier simply must be empty, then perhaps a check here could ensure that? Otherwise you might have someone regenerate/redeploy their VPC only to get to the module `deploymentunit` itself only to run into issues and have to roll-back.

aws/modules/privatebastion/module.ftl
roleyfoley: Within the existing set I'd suggest `dir` as the closest tier,  messaging is more aligned with the application where dir is more admin/infra side

roleyfoley (changes requested):

roleyfoley:
@ml019 Could you please review and see what we need to do

RossMurr4y:
All the "Type" properties should be updated to "Types" now.

roleyfoley:
Closing with the idea of converting bastion to a module

ml019:
On review, this change was about converting bastion to a module - it was stalled on deciding whether to replace the masterdata bastion with a module that could then be disabled where not needed as part of a transition to not having it in the masterdata at all. 

So I think the actions needed are
1. refactor this to be a module defined in the aws plugin
2. add an input seeder to the cmdb stage for the aws plugin to add this module by default
3. remove the bastion configuration from the masterdata
4. anywhere we actually want the seeder, we change the solution to explicitly include it - which will just overwrite was is provided form the aws loader
5. once everything is cut across, remove the seeder from the aws
6. remove any solution config which was added to disable the bstion which was configured by default

RossMurr4y (approved):
Approved and happy to merge.

@ml019 where did you land on which tier should be the default? That appears to be the only outstanding comment.

ml019:
For the approach proposed, to minimise impacts I'll use the same tier as used in the masterdata.

PR #172, opened 2020-11-21
feat: autoscale replacement updates

roleyfoley:
## Description
Implements https://github.com/hamlet-io/engine/pull/1471  and also adds process suspension during cloud formation updates 

## Motivation and Context
When cloudformation applies updates to an autoscaling group it will scale the cluster to meet the configuration parameters in the template ( min, max desired ) If other scaling events are triggered during the update the cfn stack becomes stuck in the UPDATE_IN_PROGRESS stage until it times out ( around 1-2 hours ). This is really difficult to balance as the roll back takes the same amount of time 

Suspending services which can cause instance count changes during cloudformation updates is the recommended approach to fixing this 
https://aws.amazon.com/premiumsupport/knowledge-center/auto-scaling-group-rolling-updates/ 

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/services/ec2/resource.ftl
+++ b/aws/services/ec2/resource.ftl
@@ -882,7 +882,15 @@
                 "AutoScalingRollingUpdate" : {
                     "WaitOnResourceSignals" : autoScalingConfig.WaitForSignal,
                     "MinInstancesInService" : autoscalingMinUpdateInstances,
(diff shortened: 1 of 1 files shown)


ml019 (approved):
No issues

PR #180, opened 2020-11-24
refactor(consolidatelogs): remove logwatcher support

roleyfoley:
## Description
Removes the logwatcher support from the consolidatelogs module

## Motivation and Context
We currently have two supported methods for integrating logs with the data feed component 
- Pull Approach - On the data feed you can use the LogWatchers configuration to link to components you want to collect logs from. In this approach the datafeed component needs a link to all components that it wants logs for 
- Push Approach - Using a LogProfile each component creates the log subscription as part of logging components deployment. In this approach you just need to create a logprofile and use this profile for the components which need to log to this datafeed 

From a module perspective the push approach is far more maintainable than the pull approach. 

## How Has This Been Tested?
TEsted locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/modules/consolidatelogs/module.ftl
+++ b/aws/modules/consolidatelogs/module.ftl
@@ -45,7 +45,6 @@
     lambdaSourceHash
     tier]
 
-    [@debug message="Entering Module: consolidate-logs" context=layerActiveData enabled=false /]
 
     [#local lambdaName = formatName(namePrefix + "cwlogslambda")]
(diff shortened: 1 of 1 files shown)


ml019 (approved):
Pull vs Push is often dictated by what needs to be defined as part of the component. IF the subscription can be created as part of the source object then push makes sense. If the target (datafeed) needs some config, then pull is the only option. 

In this case, push sounds the better approach because remove of a component will remove its log subscription as well.

PR #181, opened 2020-11-24
fix: bastion eip subset

### LINKED ISSUE #179: eip subset not including IP address for bastion [BUG]
## Expected Behaviour
If using an eip subset, it should include any eip required by the bastion component,

## Current Behaviour
Currently, the logic of the bastion setup routine results in the eip not being included. This is because it is conditional on the `publicRouteTable` variable being set, and it in turn is onot set if the eip subset is used.

## Possible Solution
Not sure if the subset check is really required around the code that checks for a public route table? https://github.com/hamlet-io/engine-plugin-aws/blob/ff0e09b795249e2c7745e9ab4aa8849f75a01eca/aws/components/bastion/setup.ftl#L44

## Steps to Reproduce

1. Run the segment eip unit for a project where the bastion is in a public subnet.
2. The output will not include configuration for an eip for the bastion.
3.
4.

ml019:
## Description
Ensure the eip is created as part of an eip subset.

## Motivation and Context
The logic to detect a public network (and hence the need for an eip), did not take into account an eip subset pass. It now explicitly checks for this.

Fixes #179.

## How Has This Been Tested?
Local template generation

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Followup Actions
- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/components/bastion/setup.ftl
+++ b/aws/components/bastion/setup.ftl
@@ -41,15 +41,13 @@
             [#break]
     [/#switch]
 
-    [#if deploymentSubsetRequired("bastion", true) ]
+    [#if deploymentSubsetRequired("eip", false) || deploymentSubsetRequired("bastion", true) ]
         [#local occurrenceNetwork = getOccurrenceNetwork(occurrence) ]
         [#local networkLink = occurrenceNetwork.Link!{} ]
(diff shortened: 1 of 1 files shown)


ml019:
I need to merge this to sort some deployment issues with a client project but would appreciate you checking it for any circumstance I've missed @roleyfoley

PR #182, opened 2020-11-26
fix(baseline): use s3 encryption for opsdata

roleyfoley:
## Description
Set the default encryption service for the opsdata bucket to use the S3 SSE encryption service instead of the KMS - CMK method 

## Motivation and Context
The ops data bucket is our logging destination for Load balancer logs which do not support the KMS-CMK encryption method for bucket encryption. They only support the S3 SSE method. Each object can define its own encryption process and if no default method is set load balancers still encrypt data using S3 SSE. This is a known issue with AWS and no know fix 

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/inputsources/shared/masterdata.ftl
+++ b/aws/inputsources/shared/masterdata.ftl
@@ -1009,6 +1009,9 @@
             "DataBuckets": {
               "opsdata": {
                 "Role": "operations",
+                "Encryption" : {
+                  "EncryptionSource" : "LocalService"
+                },
                 "Lifecycles": {
                   "awslogs": {
                     "Prefix": "AWSLogs",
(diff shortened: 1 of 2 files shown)


ml019 (commented):
In the order of things, will the baseline profile actually override master data?

roleyfoley:
> In the order of things, will the baseline profile actually override master data?

The baseline profile is only used by other components to determine the appropriate components in the baseline to use ( different components can potentially have their own opsdata bucket or cmk )  so this shouldn't impact that processing 

Also just added a default disable of encryption to work around the isPresent check which we use to determine encryption

ml019 (approved):
Will turn encryption off on anything that was deployed while the fix wasn't in place.

PR #184, opened 2020-11-30
fix: set nat gateway priority for mgmt contract

roleyfoley:
## Description
Minor fix and refactor for the components in the aws master data 
- Updates the NAT Gateway deployment priority to ensure its deployed after the igw 
- Uses the new deployment:Unit formatting to start the migration to the new naming format

## Motivation and Context
When using management contract deployments the unit ordering is defined as part of the solution. Most components use their component priority default, but igw and nat are both gateways, since the nat depends on the igw we need to set a specific priority for the nat. A the moment the deployment run fails when running through the segment components

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/inputsources/shared/masterdata.ftl
+++ b/aws/inputsources/shared/masterdata.ftl
@@ -1002,10 +1002,8 @@
           "Enabled": false
         },
         "baseline": {
-          "DeploymentUnits": [
-            "baseline"
-          ],
           "baseline": {
+            "deployment:Unit" : "baseline",
             "DataBuckets": {
               "opsdata": {
                 "Role": "operations",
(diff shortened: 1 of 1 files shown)


ml019:
The current logic allows for a unit to be included in more than one deployment unit. Are we now removing that? Should the value still be an array?

roleyfoley:
> The current logic allows for a unit to be included in more than one deployment unit. Are we now removing that? Should the value still be an array?

The new format moved to a single deployment unit for an occurrence in a previous PR. The main reason was that I couldn't see a reason why you would want an occurrence to belong to different deployment units. How does it decide where the resource is created? wouldn't you get the same resources deployed twice?

ml019:
The use case was more being able to identify groups of units for different purposes. Its cleaner to set it to one for deployment purposes.

If we find a use case where we need to group units differently, Probably better to manage that as a separate thing to deployment units.

ml019 (approved):

roleyfoley:
> The use case was more being able to identify groups of units for different purposes. Its cleaner to set it to one for deployment purposes.
> 
> If we find a use case where we need to group units differently, Probably better to manage that as a separate thing to deployment units.

Maybe with deployment groups?

PR #185, opened 2020-11-30
fix(ecs): handle scale in protection during updates

roleyfoley:
## Description
- Force cluster replacement on update when managed termination is enabled for EC2 instances
- Adds an empty prologue script to try and flush out the current prolgue scripts which have been saved to cmdb's at the moment they are executed if present and using the old CMDB format 

## Motivation and Context
When using managed scaling for ECS ( which is part of the capacity provider service ) ec2 instances must have scale in protection enabled, this ensures that  instances which are running active tasks aren't included aren't terminated. However when scale in protection is enabled it is also applied during cloudformation updates which trigger scaling activities themsevles. Since they wait for the desired count to reach whats specified in the cloudformation update this leaves the instances in a hung state. 

So instead we want to deploy a new cluster, confirm that this is ok and then remove the old cluster. This will impact the running containers but only during updates to the cloudformation stacks

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/ecs/setup.ftl
+++ b/aws/components/ecs/setup.ftl
@@ -1,6 +1,6 @@
 [#ftl]
 [#macro aws_ecs_cf_deployment_generationcontract_solution occurrence ]
-    [@addDefaultGenerationContract subsets=[ "template" ] /]
+    [@addDefaultGenerationContract subsets=[ "prologue", "template" ] /]
 [/#macro]
 
 [#macro aws_ecs_cf_deployment_solution occurrence ]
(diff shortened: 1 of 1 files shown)


ml019 (approved):
Quite a detailed process so will assume specifics are correct. Overall looks ok.

Flushing of now unused prologue is an interesting way to remove old files. I would have thought the current comparison logic for outputs would remove any old files if none are now generated.

PR #189, opened 2020-12-02
feat(secretstore): secrets manager support

roleyfoley:
## Description
Initial support in aws for the secrets store component 
- Allows components to use the secret store to define generated secrets for system credentials ( root passwords for managed services ) 
- Adds a dummy implementation of a component for the secrets manager secret store which doesn't require any infrastructure to actually create secrets 
- Ref support for resolving secrets in cloudformation templates 

## Motivation and Context
The primary motivation for this PR is to support the generation of admin passwords which are required for services like RDS or amazonMQ which require an initial root user to be defined in the cloudformation template. We currently set a generated password as an epilogue script. which does lead to an interruption to the service while the password is reset 

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/secretstore/id.ftl
+++ b/aws/components/secretstore/id.ftl
@@ -88,7 +88,7 @@
                         [
                             r'case ${STACK_OPERATION} in',
                             r'  create|update)',
-                            r'    info "Saving secret to CMDB",
+                            r'    info "Saving secret to CMDB"',
                             r'    secret_arn="$(get_cloudformation_stack_output "' + regionId + r'" ' + r' "${STACK_NAME}" ' + secretId + r' "ref" || return $?)"',
                             r'    secret_content="$(aws --region "' + regionId + r'" --output text secretsmanager get-secret-value --secret-id "${secret_arn}" --query "SecretString" || return $?)"',
                             r'    secret_value="$( echo "${secret_content}" | jq -r "' + secretKeyPath + r'")"',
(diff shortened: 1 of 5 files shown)


aws/components/secretstore/id.ftl
```diff
@@ -0,0 +1,86 @@
+[#ftl]
+
+[@addResourceGroupInformation
+    type=SECRETSTORE_COMPONENT_TYPE
+    attributes=[
+        {
+            "Names": "Engine",
+            "Values" : [ "secretsmanager" ],
+            "Default" : "secretsmanager"
```
RossMurr4y, line 9: As the `shared` provider doesn't grant the value "secretsmanager" I believe this new value will resolve in-engine to `aws:secretsmanager`. 

The default values aren't namespaced by the engine, so that would mean that the default value must be namespaced manually.

We could update the engine to do that for us but it would prevent a provider defaulting to a value that was initially defined by the shared provider so seems best to leave as it is. Hard to document that though...

RossMurr4y (commented):
Nice and tidy, looks great 👍

roleyfoley:
Updated to include saving a KMS encrypted version of the secret text  for generated passwords. If the secret is not generated then the whole secret is saved. This provides support for existing components

aws/services/secretsmanager/resource.ftl
```diff
@@ -0,0 +1,111 @@
...
+            requirements.RequireAllIncludedTypes
+        )
+    ]
+[/#function]
+
+[#function getSecretsManagerSecretGenerationPolicy
+        passwordLength
+        generateStringKey
+        secretTemplate
+
```
RossMurr4y, line 50:
```suggestion

```

RossMurr4y (approved):

PR #199, opened 2020-12-30
feat: Message Transfer Agent component

ml019:
## Description
Implements https://github.com/hamlet-io/engine/issues/1499

First implementation of inbound and outbound MTAs.

The outbound implementation mainly provides link and attribute support to permit sending of emails.

The inbound implementation supports rulesets and IP filters. S3, lambda and drop actions are supported.

Because inbound email support is only available in limited regions, the implementation currently assumes that if S3 encryption is required, AES will be used. KMS is more problematic as the framework support for sharing KMS keys is not yet complete.

## Motivation and Context
Simplify management of email capabilities for solutions.

## How Has This Been Tested?
Local template generation and customer deployment

## Types of changes

## Followup Actions
- [x] Depends on https://github.com/hamlet-io/engine/pull/1504

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/components/baseline/setup.ftl
+++ b/aws/components/baseline/setup.ftl
@@ -423,6 +423,22 @@
                                         "",
                                         true,
                                         "Allow CloudFront Flow Logs to use the key"
+                                    ),
+                                    getPolicyStatement(
+                                        [
+                                            "kms:Encrypt",
+                                            "kms:Decrypt",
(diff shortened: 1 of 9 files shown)


ml019:
Turns out current implementation does support use of the baseline kms key. The point to note is that it is an AWS specific attribute so needs to be included in the solution as `aws:Encryption`

PR #205, opened 2021-01-09
feat(cdn): add support for external service origins

roleyfoley:
## Description
Adds support for using an external service as a CDN origin. The external service must have 2 attributes defined, FQDN which sets the host name of the origin endpoint and PATH which can be used to define an origin path override 

## Motivation and Context
This allows for external URLs to be routed through a CDN or to allow for the creation of dummy origins used in combination with Lambda@Edge functions

## How Has This Been Tested?
Tested locally

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.

### DIFF
--- a/aws/components/cdn/setup.ftl
+++ b/aws/components/cdn/setup.ftl
@@ -355,7 +355,7 @@
 
                 [#local originHostName = originLinkTargetAttributes["FQDN"]!"HamletFatal: Could not find FQDN Attribute on external service" ]
 
-                [#local path = originLinkTargetAttributes["PATH"]!"HamletFatal: Could not fid PATH Attribute on external service" ]
+                [#local path = originLinkTargetAttributes["PATH"]!"HamletFatal: Could not find PATH Attribute on external service" ]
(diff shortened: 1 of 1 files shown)


aws/components/cdn/setup.ftl
```diff
@@ -350,6 +350,34 @@
                                 eventHandlers )]
                 [#local routeBehaviours += behaviour ]
                 [#break]
+
+            [#case EXTERNALSERVICE_COMPONENT_TYPE ]
+
+                [#local originHostName = originLinkTargetAttributes["FQDN"]!"HamletFatal: Could not find FQDN Attribute on external service" ]
+
+                [#local path = originLinkTargetAttributes["PATH"]!"HamletFatal: Could not fid PATH Attribute on external service" ]
```
ml019, line 358:
```suggestion
                [#local path = originLinkTargetAttributes["PATH"]!"HamletFatal: Could not find PATH Attribute on external service" ]
```

ml019 (changes requested):

PR #207, opened 2021-01-09
fix(userpool): set userpool region for multi region deployments

roleyfoley:
## Description
A fix to the creation of the userpool UI url which was using the region for a deployment instead of the components region

## Motivation and Context
This fixes an issue where you have a multi region deployment where a component links to a userpool in another region. Since the hosted UI url was using the `region` global variable, this didn't reflect where the userpool was deployed to. Instead we now use the placement region for the userpool component

## How Has This Been Tested?
Tested locally 

## Types of changes

## Followup Actions
- [x] None

## Checklist:
- [X] None of the above.


ml019 (approved):
All usage of region within a component should start using this pattern.

Issue #203, opened 2021-01-07
Extension Support for S3


roleyfoley:
## Expected Behaviour
The ability to control complex configuration of S3 Buckets with a focus on Bucket Policy which can include complex access policy 

## Current Behaviour
Bucket policy can only be defined based on link based policies at the moment

## Possible Solution
Add support for extensions to the s3 component which would allow for custom policy to be defined during template generation

## Context
In some cases we want to be able to set a bucket policy to allow access from all components in a given vpc, this could be used for supporting tools which don't work with aws permissions but stilll want to access the bucket

roleyfoley:
@kshychko I think the best approach here is for extensions/fragments to be added to the S3 bucket 

That requires a couple of steps 

1. defining a context as part of the component, each item in the context is a section of the component we can modify through extensions 
2. For S3 we would create a context object with a context attribute called `Policy` this would allow you to use the @Policy macro that we already have to update the S3 Policy. We also need to include the Links in the context so that we can link to the Network component and then get the VPC id from that for an S3 bucket policy 
3. Call the extensions invoke and pass the context and the occurrence 
` [#local _context = invokeExtensions( occurrence, _context )]` 
4. After invoking the Extensions you then need to take the Policy statements and add to the bucket policy as part of the component processing

roleyfoley:
All done and confirmed working

PR #209, opened 2021-01-10
Feat s3 extensions support

kshychko:
## Description

Add support for extensions to the s3 component which would allow for custom policy to be defined during template generation
## Motivation and Context

#203
## How Has This Been Tested?

Got it tested locally
## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [ ] None of the above.

### DIFF
--- a/aws/components/s3/setup.ftl
+++ b/aws/components/s3/setup.ftl
@@ -162,6 +162,7 @@
                                                         publicIPWhiteList)]
                         [#break]
                 [/#switch]
+
             [/#if]
         [/#list]
     [/#list]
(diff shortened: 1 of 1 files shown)


aws/components/s3/setup.ftl
```diff
@@ -162,6 +162,56 @@
                                                         publicIPWhiteList)]
                         [#break]
                 [/#switch]
+
+                [#local occurrenceNetwork = getOccurrenceNetwork(occurrence) ]
+                [#local networkLink = occurrenceNetwork.Link!{} ]
+                [#if networkLink?has_content ]
```
ml019, line 168: A comment explaining why it is necessary to add vpc access separate from the section that proceeds this would be useful I think for maintainers.

roleyfoley: I think it might be better to make this an extension, now that multiple extensions can be applied to a bucket this would suit that pattern and remove the dependency on S3 to know about the network

kshychko:
@ml019 , @roleyfoley I force-pushed the branch. Please check the related one in the engine repo - https://github.com/hamlet-io/engine/pull/1510

roleyfoley (approved):
Awesome work!

PR #246, opened 2021-03-09
refactor: bastion ipaddressgroups on component def

RossMurr4y:
## Description

Previously in order to configure ip access to a bastion
the configuration was applied via the segment object.

This change aligns the bastion component with other IP
address group implementations on the component in the
solution.

Existing functionality is maintained, but deemed deprecated

## Motivation and Context

Closes hamlet-io/engine#1559

## How Has This Been Tested?

Local template generation

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [x] None of the above.

### DIFF
--- a/aws/components/bastion/setup.ftl
+++ b/aws/components/bastion/setup.ftl
@@ -260,7 +260,7 @@
                         "Ports" : [ "ssh" ],
                         "IPAddressGroups" :
                             sshEnabled?then(
-                                (segmentObject.Bastion.IPAddressGroups)!(segmentObject.IPAddressGroups)![],
+                                (solution.IPAddressGroups)!(segmentObject.Bastion.IPAddressGroups)!(segmentObject.IPAddressGroups)![],
                                 []
                             ),
                         "Description" : "Bastion Access Groups"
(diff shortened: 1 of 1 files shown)


aws/components/bastion/setup.ftl
```diff
@@ -249,7 +249,7 @@
                         "Ports" : [ "ssh" ],
                         "IPAddressGroups" :
                             sshEnabled?then(
-                                (segmentObject.Bastion.IPAddressGroups)!(segmentObject.IPAddressGroups)![],
+                                (solution.IPAddressGroups)!(segmentObject.Bastion.IPAddressGroups)!(segmentObject.IPAddressGroups)![],
```
roleyfoley, line 252: Might be better to merge these rather than replace, otherwise migrations could be quite confusing. The UNIQUE behaviour on the merge is probably best

roleyfoley (changes requested):

roleyfoley:
@RossMurr4y Could you take a look

roleyfoley (approved):

Issue #257, opened 2021-03-22
SQS Support for FIFO Queues 


roleyfoley:
## Expected Behaviour

I can create an sqs queue using hamlet with FIFO ( first in first out ) ordering enabled on the queue.

## Current Behaviour

Using hamlet and its generated template there is no support for configuring the fifo support option and the default action is to create a standard queue which doesn't guarantee ordering 

## Possible Solution

Add support in the SQS cloudformation resource to enable set the FifoQueue property if required 
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sqs-queues.html#aws-sqs-queue-fifoqueue 

## Context

Ordering of work items can be critical in some applications and the FIFO support is designed to implement this functionality

PR #255, opened 2021-03-21
refactor: command line and masterdata access

ml019:
## Description
Replace direct access to the command line options with use of accessor functions. This will then permit the implementation of command line options to be refactored to the new input source approach without affecting this code.

Switch to using the input source process to determine the masterdata available.

Update the AWS masterdata refresh process to update the regions.json file in the inputseeder.

## Motivation and Context
Align with https://github.com/hamlet-io/engine/pull/1552

## How Has This Been Tested?
- local template generation
- local execution of the aws test suite

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Refactor (non-breaking change which improves the structure or operation of the implementation)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Followup Actions

- [x] None

## Checklist:

- [ ] My change requires a change to the [documentation](https://github.com/hamlet-io/docs).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] None of the above.

PR #258, opened 2021-03-22
feat(sqs): ordering configuration for queues

roleyfoley:
## Description

Adds support for enabling FIFO queues on SQS queues

- Adds the basic configuration parameter to the CFN resources
- Changing from standard queue to FIFO requires replacement and named
queues must include renaming. So added replace alternatives and naming to handle replacing when there is only the queue resource in the template ( can't remove it since that can make the template empty ) 
- Truncates the queue name to ensure that the special fifo suffix .fifo is always included in the name

## Motivation and Context

Add support for handling different queue ordering processes

closes https://github.com/hamlet-io/engine-plugin-aws/issues/257

## How Has This Been Tested?

Tested locally and on test deployment

## Types of changes

## Followup Actions

- [x] None

## Checklist:

- [X] None of the above.

### DIFF
--- a/aws/components/sqs/setup.ftl
+++ b/aws/components/sqs/setup.ftl
@@ -24,20 +24,6 @@
         [#local sqsPolicyId = resources["queuePolicy"].Id ]
         [#local queuePolicyStatements = [] ]
 
-        [#if dlqRequired ]
-            [#local dlqId = resources["dlq"].Id ]
-            [#local dlqName = resources["dlq"].Name ]
-
-            [#local queueIds += [ dlqId ]]
-            [@createSQSQueue
-                id=dlqId
-                name=dlqName
-                retention=1209600
-                receiveWait=20
-            /]
-        [/#if]
-
-
         [#local fifoQueue = false]
 
         [#switch solution.Ordering ]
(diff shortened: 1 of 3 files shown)


aws/components/sqs/state.ftl
```diff
@@ -36,6 +36,28 @@
         [#local dlqId = formatDependentResourceId(AWS_SQS_RESOURCE_TYPE, id, "dlq") ]
         [#local dlqName = formatName(name, "dlq")]
 
+        [#-- override the Id ane Name for replacement --]
+        [#if ((commandLineOptions.Deployment.Unit.Alternative)!"") == "replace1" ]
```
ml019, line 40: This will break https://github.com/hamlet-io/engine-plugin-aws/pull/255 - needs to use `getDeploymentUnitAlternative()` instead

ml019 (commented):
Looks ok but if needed soon, then I'll need to rebase and update PR #255

aws/components/sqs/setup.ftl
```diff
@@ -48,13 +85,43 @@
...
 
         [#list solution.Alerts?values as alert ]
 
             [#local monitoredResources = getCWMonitoredResources(core.Id, resources, alert.Resource)]
             [#list monitoredResources as name,monitoredResource ]
 
+                [#local resourceDimensions = []]
+                [#if ((commandLineOptions.Deployment.Unit.Alternative)!"") == "replace1" ]
+
+                    [@debug message="SQSMonResource" context=monitoredResource enabled=true /]
```
RossMurr4y, line 99: Does this always need to be on?

RossMurr4y (dismissed):

roleyfoley:
Rebased on latest changes and removed debug statement

RossMurr4y (approved):

PR #271, opened 2021-04-13
feat: aws compute task implementations

roleyfoley:
## Intent of Change

## Description

- Adds compute task definitions and extension implementations to replace the existing InitConfig scripts that we currently use in the AWS provider. 
- Adds  a new task which installs and starts the SSM agent. This is required to make SSM session manager available for ECS hosts 
- Adds a standard userdata script which calls cfn-init and uses cfn-signal to notify cloudformation of the cfn-init status
- Includes a fix to this script to ensure that the script doesn't exit at any failure to allow for quick feedback to cloudformation on the failure of an instance

## Motivation and Context

Mostly covered in https://github.com/hamlet-io/engine/pull/1629  but the idea is to allow for pluggable configuration of instances used as part of components. Extensions also allow for greater portability between different components and allow for users to build their own implementations of tasks as required. 

## How Has This Been Tested?

Tested locally

## Followup Actions

- [x] Requries: https://github.com/hamlet-io/engine/pull/1629

### DIFF
--- a/aws/extensions/computetask_linux_hamletenv/extension.ftl
+++ b/aws/extensions/computetask_linux_hamletenv/extension.ftl
@@ -1,0 +1,85 @@
+[#ftl]
+
+[@addExtension
+    id="computetask_linux_hamletenv"
+    aliases=[
+        "_computetask_linux_hamletenv"
+    ]
+    description=[
+        "Uses the shared profile configuration to set default environment varaibles"
+    ]
+    supportedTypes=[
+        EC2_COMPONENT_TYPE,
+        ECS_COMPONENT_TYPE,
+        COMPUTECLUSTER_COMPONENT_TYPE,
+        BASTION_COMPONENT_TYPE
+    ]
+    scopes=[
+        COMPUTETASK_EXTENSION_SCOPE
+    ]
+/]
+
+[#macro shared_extension_computetask_linux_hamletenv_deployment_computetask occurrence ]
+
+    [#local baselineLinks = _context.BaselineLinks]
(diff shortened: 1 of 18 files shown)


aws/extensions/computetask_linux_hamletenv/extension.ftl
```diff
@@ -0,0 +1,85 @@
...
+    [#local baselineComponentIds = getBaselineComponentIds(baselineLinks)]
+    [#local operationsBucket = getExistingReference(baselineComponentIds["OpsData"]) ]
+    [#local dataBucket = getExistingReference(baselineComponentIds["AppData"])]
+    [#local envVariables = getFinalEnvironment(occurrence, _context ).Environment ]
+
+    [#local role = (occurrence.Configuration.Settings.Product["Role"].Value)!""]
+
+    [#local envContent = [
+        r'# Set environment variables from hamlet configuration',
+        r'export cot_request="'       + getCLORequestReference()        + '"',
```
ml019, line 34: Time to switch to `hamlet_*` from `cot_*`?

ml019 (commented):
Some serious reorg. Given that, time to get rid of cot in favour of hamlet?

aws/extensions/computetask_linux_hamletenv/extension.ftl
roleyfoley: Done, I don't think these ones are actually used anymore, but didn't want to change to much of the functionality while also moving everything around

ml019 (approved):

PR #276, opened 2021-04-15
feat: awslinux2 support for ec2 instances

roleyfoley:
## Intent of Change

## Description

- Adds a deployment profile `_awslinux2` will uses the latest aws linux 2 AMI through SSM parameter store and set OS configuration for extensions to use. This can be applied either on components  or layers based on user requirements to migrate ec2 based components to awslinux2
- Updates extensions to support some of the differences between awslinux and awslinux2
- Fixes a bug in the storage setup which was not including block devices

The ECS storage profile has been updated to reflect the update to the AMI storage which now hosts everything on the root volume. If the device is specified in the block devices with a size larger than the ami on initial boot the file system will be resized to fit the size of the volume which includes the docker storage. 

## Motivation and Context

This allows for a transition to awslinux 2 from our current default of awslinux 1 which is in the process of being deprecated. The ECS optimised image no longer receives updates so the main priority is to get that introduced 

Implementation of https://github.com/hamlet-io/engine/issues/1488 
closes: https://github.com/hamlet-io/engine/issues/171

## How Has This Been Tested?

Tested locally 

## Followup Actions

- Requires: https://github.com/hamlet-io/engine/pull/1642

### DIFF
--- a/aws/extensions/computetask_awslinux_ecs/extension.ftl
+++ b/aws/extensions/computetask_awslinux_ecs/extension.ftl
@@ -28,6 +28,12 @@
 
     [#local dockerUsersEnv = "" ]
 
+    [#local commands = {}]
+    [#local services = {}]
+
+    [#local solution = occurrence.Configuration.Solution ]
+    [#local operatingSystem = solution.ComputeInstance.OperatingSystem]
+
     [#local userInit = {}]
     [#if dockerUsers?has_content ]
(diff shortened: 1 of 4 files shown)


roleyfoley:
With the move to use SSM Parameter store for the AMI Source the image Id won't be listed in the template. Instead the Stack outputs show 

```

RossMurr4y (approved):

PR #277, opened 2021-04-16
feat(ec2): adds support for autoscale lifecycles

roleyfoley:
## Intent of Change

## Description

- Adds support for autoscale lifecycle hooks focussing on the startup process. This is similar to CFN init but handles the startup instances as part of scaling activities. The instance will be started and if a failure occurs or 10 minute timeout the scaling event will terminate the instance and start again
- Adds a default lifecycle hook for startup events with the ability to override that if required.
- Adds IAM policy to allow the instances to query the ASG they belong to and send the event

## Motivation and Context

The primary motivation is to ensure that instances startup correctly with the configuration that has been given to them via cloud tasks. 

## How Has This Been Tested?
Tested locally 

## Followup Actions

- [x] None

### DIFF
--- a/aws/components/bastion/id.ftl
+++ b/aws/components/bastion/id.ftl
@@ -37,7 +37,7 @@
                         {
                             "Names" : "Extensions",
                             "Default" : [
-                                "_computetask_awslinux_cfninit",
+                                "_computetask_awslinux_cfninit_asg",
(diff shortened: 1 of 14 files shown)


ml019 (approved):
Looks like the new extension framework was designed with ASG lifecycle hooks in mind ...

PR #282, opened 2021-04-25
feat: adds support for healthchecks

roleyfoley:
## Intent of Change

## Description

Implements simple and complex healthchecks as part of the healthcheck component.

The simple check is performed by Route53 and runs distributed basic HTTP checks against a single endpoint. The check has to be run from at least 3 regions that support healthchecks and uses our standard port configuration to define the configuration of the check. A link to a destination is required for the check to work.

The simple check must be deployed to us-east-1 as cloudwatch metrics are sent to us-east-1 in a similar way to lambda@edge functions, which are run across all regions but metrics are stored in us-east-1 

The complex health check uses CloudWatch canary monitoring which uses lambda to perform scripted synthetic transactions using browser automation tools like selenium. The checks are deployed to a local region and run at a schedule

## Motivation and Context

AWS implementation of https://github.com/hamlet-io/engine/issues/1653 

This allows for transaction monitoring to be implemented as part of  hamlet solution and also allows for more complex monitoring scenarios 

## How Has This Been Tested?

Tested locally and using test suite

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1659

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/healthcheck/id.ftl
+++ b/aws/components/healthcheck/id.ftl
@@ -1,0 +1,13 @@
+[#ftl]
+
+[@addResourceGroupInformation
+    type=HEALTHCHECK_COMPONENT_TYPE
+    provider=AWS_PROVIDER
+    resourceGroup=DEFAULT_RESOURCE_GROUP
+    attributes=[]
+    services=
+        [
+            AWS_CLOUDWATCH_SERVICE,
+            AWS_ROUTE53_SERVICE
+        ]
+/]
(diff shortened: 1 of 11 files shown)


ml019:
@roleyfoley looks like the test cases don't align case wise with the implementation?

roleyfoley:
Fixed up and also fixed up some test issues with the complex healthcheck

ml019 (approved):

Issue #281, opened 2021-04-22
[BUG] Service deploying to default cluster if cluster not deployed


roleyfoley:
## Current Behaviour

When a service is deployed and the solution level ecs cluster hasn't been deployed the template generates successfully and sets the Cluster value to an empty string. 

This is accepted by Cloudformation and the service is deployed to the default ecs cluster instead of the specified cluster. When using fargate this works fine as the service and tasks can be started

## Expected Behaviour

The service or task should fail if their ecs cluster parent details can't be found

## Possible Solution

Add check for cluster setting and raise a fatal message if empty

## Steps to Reproduce

1. Create a ecs component and add a service deployed to fargate
2. Run the deployment of the service 
3. The template will generate and deploy to the default cluster

PR #284, opened 2021-04-26
feat(ecs): align with latest aws features

roleyfoley:
## Intent of Change

## Description

A collection of changes to align with the latest aws features for ECS

- Uses the capacity provider association resource to define the capacity provider and ASG link. This removes the circular dependency when provisioning the ASG and ECS cluster at the same time
- Replaces the LaunchTarget configuration with capacity providers. Services have their own capacity provider configuration based on their engine configuration as the task definition is still configured based on a particular platform ( ec2 or fargate )
- Adds circuit breaker deployment support which will roll back deployments which can't stabilise the containers due to startup failures. This removes the need to cancel failed deployments via the console
- Fixes: https://github.com/hamlet-io/engine-plugin-aws/issues/281 - an issue where tasks would deploy to the default ECS cluster if the parent component wasn't deployed

## Motivation and Context

The primary motivation in this PR was to enable capacity providers for the fargate engine in order to provide access to Fargate spot. This can significantly reduce hosting costs especially for non production environment background tasks which can easily handle failures 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1661

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
@roleyfoley looks like the test cases need to be aligned to the refactored capacity provider code?

ml019 (approved):

PR #296, opened 2021-05-07
feat(ec2): handle post tasks for ec2

roleyfoley:
## Intent of Change

## Description

- Adds a new compute task engine which is intended for post provisioning of ec2 instances after they have been created. This is useful when you need to configure resources which are dependent on the ec2 resource and won't be setup until the ec2 instance is complete
- Adds CFN wait resources which are used to provider config validation on the post provisioning stage of the ec2 instance config
- Aligns ec2 based components with the data volume vs system volume mount configuration changes which align with this one as well
- Adds replacement configuration to the ec2 instances so that updates made to the bootstrap configuration are applied on a new instance during updates. 

## Motivation and Context

This primary motivation for this is to support mounting data volumes to ec2 hosts. When creating both the ec2 instance and attaching a data volume the script due to a dependency between the volume attachment and the ec2 instance creation. 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1679

### Dependent PRs:

- None

### Consumer Actions:

- None


kshychko:
@roleyfoley , normal stack update doesn't seem to work to me, it's failing with the error that a device is not available. Although running update in stopstart mode works well.

ml019:
@roleyfoley, code looks ok but I'll wait for response to @kshychko before approving/merging.

Issue #299, opened 2021-05-10
Ec2 component support for windows


roleyfoley:
## Expected Behaviour

Using a deployment profile I can specify the image source and extensions required to deploy a microsoft windows based ec2 instance as an ec2 component

## Current Behaviour

While we support specifying your own image and extensions for compute instances a user would need to implement everything themselves without support from hamlet itself

## Possible Solution

- Add a new masterdata profile `_compute_windows2019` which is used to set the appropriate compute instance configuration for our recommended windows instance setup
- Add extensions which implement the compute tasks we include in the ec2 component based on windows requirements 

## Context

Adding support for this would highlight the hamlet flexibility in ec2 setup and how it works

roleyfoley:
@ken9ross  This ticket is to cover off the basics of the windows support with a focus on the ec2 component

For the ec2 component we have the following tasks which I've divided into groups: 

base system configuration - these tasks don't have a dependency on other components or configuration and are mostly focussed on getting the system up and running 
```
                        COMPUTE_TASK_RUN_STARTUP_CONFIG,
                        COMPUTE_TASK_AWS_CFN_SIGNAL,
                        COMPUTE_TASK_AWS_CFN_WAIT_SIGNAL,
                        COMPUTE_TASK_OS_SECURITY_PATCHING,
                        COMPUTE_TASK_AWS_CLI
```

Extensions and profiles - these tasks use information collected from either extension/fragment processing or from reference data 
```
                       COMPUTE_TASK_SYSTEM_VOLUME_MOUNTING,
                       COMPUTE_TASK_HAMLET_ENVIRONMENT_VARIABLES,
                       COMPUTE_TASK_USER_BOOTSTRAP,
                       COMPUTE_TASK_SYSTEM_LOG_FORWARDING,
                       COMPUTE_TASK_FILE_DIR_CREATION
```

Link based config - these tasks use links from the component to other components to perform different tasks 
```
                        COMPUTE_TASK_DATA_VOLUME_MOUNTING,
                        COMPUTE_TASK_AWS_LB_REGISTRATION,
                        COMPUTE_TASK_EFS_MOUNT,
                        COMPUTE_TASK_USER_ACCESS,
````

From a complexity point of few they get more complicated as things get involved. The base system configuration tasks are used across most of our ec2 based components then as they get more complicated they change for each component

The ec2 component does have special configuration to handle some extra dependency requirements but its what would be the base level of configuration.

[...]

PR #301, opened 2021-05-11
Windows instances under AWS

ken9ross:
## Intent of Change

## Description
Add support for Windows instances in AWS

## Motivation and Context
Capability building
part of #299 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- New build profile of awswin similar to awslinux2


aws/extensions/computetask_awscli/extension.ftl
```diff
@@ -26,6 +26,9 @@
 
     [#local content = {}]
     [#switch operatingSystem.Family ]
+        [#case "windows" ]
+        [#break]
+
```
roleyfoley, line 31: Are we not installing the aws cli? or is it already bundled?

aws/extensions/computetask_awswin_cfninit/extension.ftl
```diff
@@ -0,0 +1,64 @@
...
+    ]
+    supportedTypes=[
+        EC2_COMPONENT_TYPE,
+        ECS_COMPONENT_TYPE,
+        COMPUTECLUSTER_COMPONENT_TYPE,
+        BASTION_COMPONENT_TYPE
+    ]
+    scopes=[
+        COMPUTETASK_EXTENSION_SCOPE
+    ]
```
roleyfoley, line 19: This now needs to include handling of a second cfn-init pass for when we need to handle cloudformation actions that depend on the ec2 instance

roleyfoley: Actually looks like you've added another function for that. This one can be removed. We either have wait or autoscale group now

aws/extensions/computetask_awswin_ecs/extension.ftl
```diff
@@ -0,0 +1,227 @@
...
+            [#local userInit = mergeObjects(
+                                userInit,
+                                {
+                                    userName : {
+                                        "groups" : [ "docker" ],
+                                        "uid" : details.UID
+                                    }
+                                })]
+        [/#list]
+    [/#if]
```
roleyfoley, line 50: Might need to review this and see what the windows config is for user provisioning.

aws/extensions/computetask_awswin_ospatching/extension.ftl
```diff
@@ -0,0 +1,74 @@
...
+
+    [#local OSPatching = _context.InstanceOSPatching ]
+    [#local schedule = OSPatching.Schedule ]
+    [#local securityOnly = OSPatching.SecurityOnly ]
+
+    [#local updateCommand = "yum clean all && yum -y update"]
+
+<!-- Windows instances are patched by new AMIs
+    [#if OSPatching.Enabled ]
+ -->
```
roleyfoley, line 32: The awslinux instances are as well. The AMIs are released in monthly cycles I think. so could miss some updates that have been released in between. the AMI cycle

aws/inputseeders/aws/masterdata.json
```diff
@@ -946,6 +956,127 @@
                     }
                 }
             }
+        },
+        "_awswin" : {
```
roleyfoley, line 960: I think this is probably better as windows rather than awswin. awslinux is an actual distribution. Where the Windows install is really windows.

aws/inputseeders/aws/masterdata.json
```diff
@@ -946,6 +956,127 @@
...
+                                "_computetask_awswin_vpc_lb",
+                                "_computetask_awswin_notavailable"
+                            ]
+                        }
+                    },
+                    "computecluster" : {
+                        "ComputeInstance" : {
+                            "OperatingSystem" : {
+                                "Family" : "windows",
+                                "Distribution" : "awswin",
```
roleyfoley, line 998: I don't think this is a real distribution. We should reserve this for when its required.

roleyfoley (changes requested):

aws/extensions/computetask_awscli/extension.ftl
ken9ross: Doco says AWS CLI is already bundled

aws/inputseeders/aws/masterdata.json
ken9ross: Since it's not a vanilla windows install I was appropriating the term "distribution" to describe the AWS customised windows install (AWS CLI, EC2Lauch etc). It's not as if there is (or likely ever will be - would need MS to opensource Windows) a Redhat Windows and a MS Windows, unlike Linux.

The other potential "distribution" I thought about was the "preview" windows versions from AWS. I was thinking of awswin and awspwin, but they could be "preview" and "ga" distributions just as easily. Thoughts?

aws/extensions/computetask_awswin_ecs/extension.ftl
ken9ross: Sorry yes, ecs is currently junk and a placeholder for me

ml019:
@ken9ross the profile refactoring has been merged so you should be good to rebase and maybe move this to review?

aws/extensions/computetask_awswin_hamletenv/extension.ftl
```diff
@@ -66,6 +66,11 @@
         priority=2
         engine=AWS_EC2_CFN_INIT_COMPUTE_TASK_CONFIG_TYPE
         content={
+                "packages" : {
+                    "msi" : {
+                        "awscli": "https://awscli.amazonaws.com/AWSCLIV2.msi"
+                    }
```
roleyfoley, line 72: Should we move this to the awscli step? then its handled as part of the same extension?

ken9ross: Good point - done

aws/inputseeders/aws/masterdata.json
```diff
@@ -648,6 +693,23 @@
...
+            "TimeFormat": "%Y/%m/%d %H:%M:%S%Z:"
+        },
+        "/cwa/amazon-cloudwatch-agent.log": {
+            "FilePath": "/cwa/amazon-cloudwatch-agent.log",
+            "TimeFormat": "%H: %M: %S%y%b%-d"
+        },
+        "/awsssm/amazon-ssm-agent.log": {
+            "FilePath": "/awsssm/amazon-ssm-agent.log",
+            "TimeFormat": "%Y-%m-%d %H:%M:%S",
+            "MultiLinePattern": "^INFO"
```
roleyfoley, line 712: Are these the windows logs? Lest make them Windows formatted instead of something hard to follow.

aws/inputseeders/aws/masterdata.json
```diff
@@ -952,6 +1048,176 @@
...
+            "Modes" : {
+                "*" : {
+                    "ecs" : {
+                        "Profiles" : {
+                            "Storage" : "ecs_awswin",
+                            "LogFile" : "awswinlog"
+                        },
+                        "AutoScaling": {
+                            "StartupTimeout": "20M"
+                        },
```
roleyfoley, line 1062: 20 minutes for each box? seems quite long.

aws/services/cw/policy.ftl
```diff
@@ -17,7 +17,18 @@
...
+            [
+                getPolicyStatement(
+                    [
+                        "cloudwatch:PutMetricData",
+                        "ec2:DescribeTags"
+                    ],
+                    "*")
+            ],
+            []
+        )
```
roleyfoley, line 31: I don't think these are OS dependent.  the linux instances can put metric data as well. We just didn't have it setup. Probably better to make this standard permissions on cwLogs across the board

roleyfoley (changes requested):

aws/extensions/computetask_awswin_ospatching/extension.ftl
ken9ross: Not sure this is the same for Windows as patches are released on "Patch Tuesday" (ie monthly) and the AMIs are updated with 5 working days thereafter. I'm not aware of people using automated patching on Windows servers, but I'll have a dig through Windows Update

aws/inputseeders/aws/masterdata.json
ken9ross: Yes and no. The FilePath is used to identify the windows log and also the logstream name, hence using a function to translate between them. Would it be better to create a second parameter?

aws/inputseeders/aws/masterdata.json
ken9ross: The doco lists up to 15 minutes for a box to start and 15 minutes to join a cluster. If I've read the doco aright, that's a maximum time for a response (not a minimum), correct?

aws/extensions/computetask_awswin_cwlog/extension.ftl
```diff
@@ -0,0 +1,233 @@
...
+        ECS_COMPONENT_TYPE,
+        COMPUTECLUSTER_COMPONENT_TYPE,
+        BASTION_COMPONENT_TYPE
+    ]
+    scopes=[
+        COMPUTETASK_EXTENSION_SCOPE
+    ]
+/]
+
+[#function windows_path_converter nix_path ]
```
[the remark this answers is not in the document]
ken9ross: See below comment about FilePath being used to identify the windows log and also the logstream name

aws/extensions/computetask_awswin_ospatching/extension.ftl
```diff
@@ -0,0 +1,66 @@
...
+    ]
+/]
+
+[#macro shared_extension_computetask_awswin_ospatching_deployment_computetask occurrence ]
+
+    [#local OSPatching = _context.InstanceOSPatching ]
+    [#local schedule = OSPatching.Schedule ]
+    [#local securityOnly = OSPatching.SecurityOnly ]
+
+    [#local updateCommand = "<script>rem Windows no update command - yum clean all && yum -y update</script>"]
```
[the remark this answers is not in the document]
ken9ross: nix functionality that was not translated to Windows. Little more than a memory jogger as to what wasn't translated in a comment

aws/services/cw/policy.ftl
ken9ross: Cool, was just trying to constrain my changes to avoid breaking something out of ignorance

aws/extensions/computetask_awswin_volumemount/extension.ftl
```diff
@@ -0,0 +1,176 @@
...
+        [/#list]
+    [/#if]
+
+    [#local diskId = -1]
+    [#list volumes as id,volume ]
+
+        [#local deviceId = ""]
+        [#local osMount = ""]
+
+        [#local diskId = diskId + 1]
```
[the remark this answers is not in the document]
ken9ross: Joys of learning a new language - missing the obvious

But a fix in volume mounting, makes this moot - it needs to be restricted to enabled volumes, so there was an introduced bug - now fixed

aws/extensions/computetask_awswin_ospatching/extension.ftl
```diff
@@ -0,0 +1,66 @@
...
+                    },
+                    {
+                        "DailyUpdates" : {
+                            "command" : '<script>echo \"${schedule} ${updateCommand} >> c:\\ProgramData\\Hamlet\\Logs\\update.log\" >crontab.txt && crontab crontab.txt</script>',
+                            "ignoreErrors" : false
+                        }
+                    }
+                )
+            }]
+    [#else]
```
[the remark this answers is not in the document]
ken9ross: Yes. I left it in as a placeholder for possible taskscheduler integrations if required

aws/extensions/computetask_awswin_ospatching/extension.ftl
roleyfoley: Lets leave it for now and instead add a warning or fatal message that if OSPatching is Enabled on the solution configuration let the user know at template generation time that the standard OS patching rules apply. 

If someone wants to enable patching they can add their own extension

aws/extensions/computetask_awswin_ospatching/extension.ftl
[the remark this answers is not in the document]
roleyfoley: Lets take it out its confusing for someone picking this up if they need to work with it. At the moment it just looks like a bug.

aws/inputseeders/aws/masterdata.json
roleyfoley: Might be better to include the full path. I.e C:/ etc 
Otherwise this can't be customised by people. What would happen if I want to log something on the D: Drive in this case? 

Looking at https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-logs-logstream.html 
`:` is not allowed as part of the name so we could just do a replace when setting the logstream to remove that 


C/ProgramFiles/ is pretty understandable

aws/inputseeders/aws/masterdata.json
roleyfoley: Its more the impact on a system if the box fails to start correctly. 20 minutes to wait for an instance to timeout when it didn't start properly is a pretty big wait during scaling activities.

aws/services/cw/policy.ftl
roleyfoley: Its just that this update isn't conditional on if you are using windows or not. Its more conditional on if you are using metrics as part of your cloudwatch agent configuration. Along with that we probably need to apply some conditions to the PutMetricData action as it's a write action. If I had multiple products in this account this permission could put data to any metric

aws/extensions/computetask_awswin_ospatching/extension.ftl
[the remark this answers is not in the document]
ken9ross: Cool, done

aws/extensions/computetask_awswin_ospatching/extension.ftl
ken9ross: Cool, done

aws/extensions/computetask_awswin_ospatching/extension.ftl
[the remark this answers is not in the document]
ken9ross: Removed

aws/inputseeders/aws/masterdata.json
ken9ross: If an ECS has a data volume it needs to do an NTFS format plus doco warns up to 15 minutes to join the cluster and can be successful

I couldn't locate how long nix was expected to take to join a cluster, but given the above it didn't seem too much to double it to 20M. Thoughts?

The risk point is an ECS that doesn't positively succeed nor positively fail but doesn't start correctly, correct?

roleyfoley: Ok sounds reasonable then. Just wanted to make sure that some of the dev work wasn't bleeding into the default setup.

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
roleyfoley: How will this affect enabling and disabling volumes? Would it still be better to use the index value and just have gaps in the Id format?

aws/services/cw/policy.ftl
ken9ross: I hadn't considered products. Logically I should be restricting it base on product and environment, but I don't see how to set that through the permissions ARN. Can you point me in the right direction?

aws/inputseeders/aws/masterdata.json
ken9ross: Cool, done

aws/extensions/computetask_awswin_ospatching/extension.ftl
ken9ross: I'm getting this warning even though I've specified OSPatching/Enabled:false in masterdata for the _awswin DeploymentProfile. Any advice on what I've done wrong?

aws/services/cw/policy.ftl
roleyfoley: This is the one you want 
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/iam-cw-condition-keys-namespace.html 

So either there is a namespace created for each product/env/segment or we limit it to a single namespace where all ec2 instances log their data under different dimensions

aws/inputseeders/aws/masterdata.json
roleyfoley: not quite what I was thinking 

For the FilePath It should be the real OS path. We just do the removal of the : when the logstream property is set
Then the user understands and enters a real path and we handle the weirdness of AWS underneath. With the configuration as it is we are now showing the weirdness to users and making them have to understand it

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
ken9ross: It doesn't, unless I'm misunderstanding how that works. Windows doesn't use the volume names but the incremental number of the volume (based on the order of the enabled volumes, so I have to make sure the numbering aligns is all. eg
```
                "Volumes": {
                    "root" : {
                        "Device" : "/dev/sda1",
                        "Size" : 40
                    },
                    "userdata" : {
                        "Enabled" : false,
                        "Device" : "xvdd",
                        "Size" : 50,
                        "MountPath" : "D"
                    },
                    "data": {
                        "Enabled" : true,
                        "Device": "xvdz",
                        "Size": 10,
                        "MountPath" : "c:\\esdata"
                    }
                }
```
Device /dev/sda1 is Disk0 and device xvdz is Disk1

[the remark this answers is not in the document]
roleyfoley: So if thats defined by the Id and I disable and enable a volume on one of my deployments what would happen?

aws/services/cw/policy.ftl
ken9ross: I think the piece of the puzzle that I'm missing is hidden in that last sentence. As product/env/segment are constructs of ours, how do they relate to ARNs (I'm assuming an ARN is a unique reference, correct?)

aws/inputseeders/aws/masterdata.json
ken9ross: Cool, done

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
ken9ross: Then during that execution (for example) device /dev/sda1 is Disk0, device xvdd would become Disk1 and device xvdz would be removed (assuming you set xvdd to enabled and xvdz to disabled). Disk0 and disk1 are used by diskpart during the userdata run to connect the disks when the EC2 spins up.

Does that answer your question or am I missing something?

aws/services/cw/policy.ftl
roleyfoley: So its within the configuration of the agent. 

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html 

within the Metrics configuration section there is the definiton of the namespace. This is the top level grouping that is used to store CloudWatch Metrics under 

So we could create a namespace that is based on the name of the component or the path of the component name. Then all instance metrics would be stored under that using the instance Id as a dimension which allows for metrics to be stored under different granularities 

Then when creating the IAM policy you would provide the Metric namespace that the role has permissions to write into

aws/extensions/computetask_awswin_cwlog/extension.ftl
```diff
@@ -0,0 +1,245 @@
...
+                                r'                "measurement": [ ',
+                                r'                    "% Committed Bytes In Use" ',
+                                r'                ], ',
+                                r'                "metrics_collection_interval": 60 ',
+                                r'            }, ',
+                                r'            "statsd": { ',
+                                r'                "metrics_aggregation_interval": 60, ',
+                                r'                "metrics_collection_interval": 10, ',
+                                r'                "service_address": ":8125" ',
+                                r'            } ',
```
roleyfoley, line 221: We shouldn't need statsd by default this is a service endpoint that applications can send metrics to. I'd call this an advanced feature and adding it by default might be a bir far

aws/extensions/computetask_awswin_cwlog/extension.ftl
```diff
@@ -0,0 +1,245 @@
...
+                                r'                        "log_stream_name": "{instance_id}/security" ',
+                                r'                    } ',
+                                r'                ] ',
+                                r'            } ',
+                                r'        } ',
+                                r'    }, ',
+                                r'    "metrics": { ',
+                                r'        "append_dimensions": { ',
+                                r'            "AutoScalingGroupName": "${aws:AutoScalingGroupName}", ',
+                                r'            "ImageId": "${aws:ImageId}", ',
```
roleyfoley, line 197: Image Id might be  a bit far here, I think just ASG, InstanceId and Type are good dimensions

roleyfoley (changes requested):
A couple of extra things. 

Also why are there double \\ in all of the paths?

aws/extensions/computetask_awswin_cwlog/extension.ftl
ken9ross: Cool

aws/extensions/computetask_awswin_cwlog/extension.ftl
ken9ross: Cool

aws/services/cw/policy.ftl
ken9ross: Cool, thanks, done

aws/services/cw/policy.ftl
```diff
@@ -21,7 +21,24 @@
...
+                    "cloudwatch:PutMetricData"
+                ],
+                "*",
+                "",
+                logGroupName?has_content?then(
+                    {
+                        "StringEquals": {
+                            "cloudwatch:namespace": 
+                                "CWAgent"+logGroupName?keep_before_last("/")?keep_before_last("/")
+                        }
```
roleyfoley, line 33: This feels quite confusing and I wouldn't know the name straight off, since its used for alerts it needs to be easy to understand what the value will be 

- I think this would be better as its own macro that returns a separate policy specific for metric data pushing. And that has a namespace value. That way its formatted outside of the policy statement and can be included as part of the ec2 instance state potentially 
- We could also leave out creating a dedicated namespace for each component and rely on the dimensions provided for each ec2 instance. This would align with how metrics are collected. Then the permission applied would always been to the CWAgent namespace for these metrics and the dimensions can be filtered. That still limits the potential for overriding metrics but makes it easier to understand and aligns with how metrics are stored across the board

roleyfoley (changes requested):

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
roleyfoley: My question is more how flexible is the disk updates on an existing instance? Can I add and remove disks on an existing instance or will adding a new one force it to mount the disk to another location? 

I.e. if I add a disk in between root and userdata will that disk mount to the location it has been configured to use or will that use the diskId for userdata and end up on the D: drive?

aws/extensions/computetask_awswin_ospatching/extension.ftl
roleyfoley: Is this still happening or was it resolved?

aws/extensions/computetask_awswin_cwlog/extension.ftl
```diff
@@ -33,6 +33,16 @@
     [#return retval ]
 [/#function]
 
+[#function nix_path_converter win_path ]
+    [#local seqWin = win_path?split(r"\")]
+    [#if (seqWin[0])?matches('[c-zC-Z]:')]
+        [#local seqWin = [(seqWin[0])?keep_before(":")] + seqWin[1..]]
```
roleyfoley, line 39: Why not 

path?replace(":", "") 

That would remove any reference to : which is not allowed at anypoint in the logstream name?

aws/extensions/computetask_awswin_cwlog/extension.ftl
```diff
@@ -51,9 +61,9 @@
             [#local logContent +=
                 [
                     r'{',
-                    r'    "file_path": "' + windows_path_converter(logFileDetails.FilePath) + '",',
+                    r'    "file_path": "' + (logFileDetails.FilePath)?replace(r"\",r"\\") + '",',
```
roleyfoley, line 64: I think the JSON output handling should cover the conversion and escaping. Was it not?

roleyfoley (changes requested):

aws/extensions/computetask_awswin_ospatching/extension.ftl
ken9ross: Still happening

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
ken9ross: It won't force it to another location (since the location is explicit - ie drive letter or path - and separate from the disk number), but there will be an issue with disks as Windows doesn't use raw disks (so they have to be formatted). I'll need to check that existing disks aren't clobbered during an update

aws/extensions/computetask_awswin_ospatching/extension.ftl
roleyfoley: Should now be fixed in b211b66 
It looks like the Environment Patching configuration was overriding the instance. The instance should be preferred as its the most specific

aws/extensions/computetask_awswin_cwlog/extension.ftl
roleyfoley: Fixed and moved to using native object support to handle escaping and make it easier to follow the extension 

see 4a9fa5f

aws/extensions/computetask_awswin_cwlog/extension.ftl
roleyfoley: Covered in 4a9fa5f

aws/services/cw/policy.ftl
roleyfoley: Addressed in 1bd6d75

aws/extensions/computetask_awswin_volumemount/extension.ftl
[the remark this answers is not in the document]
roleyfoley: @ken9ross do we have an answer on your checks on the disks?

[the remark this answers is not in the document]
roleyfoley: Testing of this has shown that the instances are replaced when disks are added, so this means that there isn't a risk of disks being remapped and causing data to be overriden. I think we can call this resolved and work through it if it becomes an issue

roleyfoley (approved):
Great Work @ken9ross awesome to test out how flexible our approach is and get a good result from it

PR #340, opened 2021-06-22
refactor: volume handling on ec2 instances

### LINKED ISSUE #338: Question: About handling of Enabled=False in volumemount
It appears that if a mount is defined in masterdata.json as Enabled=false
```
            "ECS": {
                "Volumes" : {
                    "docker" : {
                        "Device" : "/dev/xvdcz",
                        "Size" : 50
                    },
                    "dockerVolume" : {
                        "Enabled" : false,
                        "Device" : "/dev/xvdv",
                        "Size" : "50",
                        "MountPath" : "/var/lib/docker/volumes"
                    }
                }
            },
```
The volume `dockerVolume` is still created using AWS::EC2::VolumeAttachment but not mounted to `/var/lib/docker/volumes`

Is that correct? What is the purpose of doing that?

roleyfoley:
## Intent of Change

## Description

- Uses the context StorageProfile to align with its use in extensions
- Adds support for configuring volume types and Iops on block devices
- Fixes volume state handling in the volume mount extension
- Fixes types in the masterdata to align the configuration schema

## Motivation and Context

fixes #338 
This adds better handling inline with the configuration provided by the component schema. Some properties in the Storage Profile don't appear to have been implemented 

Adds enabled checks on volumes to ensure they are handled as described 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1712

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/extensions/computetask_linux_volumemount/extension.ftl
+++ b/aws/extensions/computetask_linux_volumemount/extension.ftl
@@ -33,6 +33,8 @@
     [#-- Support for data volumes as linked components to single instances --]
     [#if occurrence.Core.Type == EC2_COMPONENT_TYPE ]
 
+        [#-- These volumes are provided through DataVolumes which are an indepdent component --]
+        [#-- These are attached to instances through links and VolumeMount macro which applies the extra configuration required --]
         [#local zoneResources = occurrence.State.Resources.Zones]
         [#list zones as zone]
             [#if multiAZ || (zones[0].Id = zone.Id)]
@@ -57,6 +59,7 @@
 
                             [#local volumes += {
                                 mountId : {
+                                    "Enabled" : true,
                                     "MountPath" : volumeMount.MountPath,
                                     "Device" : volumeMount.DeviceId,
                                     "DataVolume" : true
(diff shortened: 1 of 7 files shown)


aws/extensions/computetask_linux_volumemount/extension.ftl
```diff
@@ -75,9 +78,7 @@
         [#local deviceId = ""]
         [#local osMount = ""]
 
-        [#if (volume.Enabled)!true
-                && ((volume.MountPath)!"")?has_content
-                && ((volume.Device)!"")?has_content ]
+        [#if volume.Enabled && volume.MountPath?? && volume.Device?? ]
```
ml019, line 81: semantics are slightly different. This checks that the attribute exists but could be empty. The previous code checked for an actual value.

Is that intentional?

ml019 (approved):

aws/extensions/computetask_linux_volumemount/extension.ftl
roleyfoley: Yeah this was intentional. The MountPath is an optional value to allow for creating a disk but not formatting it. 
Providing MountPath "" as a default felt weird for that action. 

Is that mount to the root or is that don't mount? leaving it as doesn't exist felt better

PR #345, opened 2021-06-24
fix(ecs): handle daemon launch mode for ec2 hosts

### LINKED ISSUE #344: [BUG] ECS daemon tasks failing to deploy with capacity providers
## Current Behaviour

When running the deployment of a ecs service running in daemon launch mode the cloudformation template fails with the following error 

```
An error occurred (InvalidParameterException) when calling the CreateService operation: Specifying a capacity provider strategy is not supported when you create a service using the DAEMON scheduling strategy.
```

## Expected Behaviour

The deployment should work and not throw an error

## Possible Solution

update the ECS setup component to always use the EC2 Launch type when the run mode is DAEMON. This is the expected behaviour since the DAEMON task should run across all ec2 instances that are part of the cluster

## Steps to Reproduce

1. Create an ecs and service component with placement.Strategy set to daemon on the service
2. run the deployment using the aws provider

roleyfoley:
## Intent of Change

## Description

- don't include capacity provider strategy on ecs services using daemon placement
- ensure ec2 launch type is set when using daemon placement

## Motivation and Context

fixes #344 which would fail when a capacity provider was included on the service definition when using DAEMON mode

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (approved):
Deceptively simple fix ...

Issue #361, opened 2021-07-14
Raise error when ecs service or task has no container definitions


roleyfoley:
## Expected Behaviour

When I create a service or task and don't include containers the hamlet engine should raise a fatal error and ask for container definitions to be added

## Current Behaviour

The service or task template is generated and its up to cloudformation to reject the change 

## Possible Solution

on the createTaskDefinition macro add a check to count the number of container definitions and raise a fatal if not above 1

## Context

This makes it easier to define and understand the requirements for the component

PR #364, opened 2021-07-20
Fatal error if no containers for service or task

ken9ross:
## Intent of Change

## Description
Raise Hamlet fatal error for task without container 

## Motivation and Context
Fixes https://github.com/hamlet-io/engine-plugin-aws/issues/361
Shift onus of test onto Hamlet

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/ecs/setup.ftl
+++ b/aws/components/ecs/setup.ftl
@@ -1624,7 +1624,8 @@
             /]
             
             [#if containers?size < 1 ]
-                [@fatal message="No container available" context=resources["task"] /]
+                [@fatal message="No container available. Add one or more containers to the following service/task" 
+                    context=resources["task"] /]
             [/#if]
 
         [/#if]
--- a/aws/components/ecs/setup.ftl
+++ b/aws/components/ecs/setup.ftl
@@ -1611,17 +1611,21 @@
                 [/#list]
             [/#list]
 
-            [@createECSTask
-                id=taskId
-                name=taskName
-                engine=engine
-                containers=containers
-                role=roleId
-                executionRole=executionRoleId
-                networkMode=networkMode
-                dependencies=dependencies
-                fixedName=solution.FixedName
-            /]
+            [#if containers?size > 0 ]
+                [@createECSTask
+                    id=taskId
+                    name=taskName
+                    engine=engine
+                    containers=containers
+                    role=roleId
+                    executionRole=executionRoleId
+                    networkMode=networkMode
+                    dependencies=dependencies
+                    fixedName=solution.FixedName
+                /]
+            [#else]
+                [@fatal message="No container to deploy services or tasks into" /]
+            [/#if]
 
         [/#if]
 
--- a/aws/components/ecs/setup.ftl
+++ b/aws/components/ecs/setup.ftl
@@ -1611,20 +1611,20 @@
                 [/#list]
             [/#list]
 
-            [#if containers?size > 0 ]
-                [@createECSTask
-                    id=taskId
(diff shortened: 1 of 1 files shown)


aws/components/ecs/setup.ftl
```diff
@@ -1611,17 +1611,21 @@
...
+                    engine=engine
+                    containers=containers
+                    role=roleId
+                    executionRole=executionRoleId
+                    networkMode=networkMode
+                    dependencies=dependencies
+                    fixedName=solution.FixedName
+                /]
+            [#else]
+                [@fatal message="No container to deploy services or tasks into" /]
```
roleyfoley, line 1627: The exception probably needs some more context.

- The containers belong to the task/service not the other way around  
- The message should be a quick message outlining what the problem was 
- The context of the message should include things like taskId 

You also don't need to make creating the task conditional. The createTemplate process will fail as a fatal exception has been added to the log messages. Making the task creation conditional means that we can't look at the fatal template during debugging. 

so might be better to just put the if conditional around the fatal message

roleyfoley (changes requested):

aws/components/ecs/setup.ftl
ken9ross: Cool

aws/components/ecs/setup.ftl
```diff
@@ -1622,6 +1622,10 @@
                 dependencies=dependencies
                 fixedName=solution.FixedName
             /]
+            
+            [#if containers?size < 1 ]
+                [@fatal message="No container available" context=resources["task"] /]
```
roleyfoley, line 1627: Is that context useful enough to be able to identify the task or service where this has happened? 
If you looked at what is provided could you fix the issue?

ken9ross: That sounds like I'm missing something obvious. I get an error message of
```
{
  "Timestamp": "2021-08-02T08:31:24.602+10:00",
  "Severity": "fatal",
  "Message": "No container available",
  "Context": {
    "Id": "ecsTaskXappXserviceecsbaseXtaskbase",
    "Name": "application-serviceecsbase-taskbase",
    "Type": "ecsTask",
    "Deployed": false
  }
}
```
With a solution file of
```
  "Tiers": {
    "app": {
      "Components": {
        "serviceecsbase": {
          "ecs": {
<snip>
            "Tasks": {
              "taskbase": {
                "Instances": {
                  "default": {
                    "deployment:Unit": "aws-task-base"
                  }
                },
<snip>
```

roleyfoley: From the exception message could I figure out what I need to do to fix the problem that has caused a fatal error to occur? 

I think it needs a suggestion on what to do or a hint on whats going on. Its also containers not container. You can have multiple containers in a task or service definition

ken9ross: Like
```
{
  "Timestamp": "2021-08-02T16:49:20.496+10:00",
  "Severity": "fatal",
  "Message": "No container available. Add one or more containers to the following service/task",
  "Context": {
    "Id": "ecsTaskXappXserviceecsbaseXservicebase",
    "Name": "application-serviceecsbase-servicebase",
    "Type": "ecsTask",
    "Deployed": false
  }
}
```
Or am I still missing the point

roleyfoley: Yeah I think that covers it.

roleyfoley (approved):

PR #365, opened 2021-07-27
Add tags for SQS

ken9ross:
## Intent of Change

## Description
Include tags for SQS components

## Motivation and Context
https://github.com/hamlet-io/engine/issues/1133

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/sqs/setup.ftl
+++ b/aws/components/sqs/setup.ftl
@@ -74,7 +74,7 @@
                 tags=
                     getOccurrenceCoreTags(
                         occurrence,
-                        formatComponentFullName(core.Tier, core.Component, dlqName))
+                        dlqName)
                 fifoQueue=fifoQueue
             /]
         [/#if]
@@ -96,7 +96,7 @@
             tags=
                 getOccurrenceCoreTags(
                     occurrence,
-                    formatComponentFullName(core.Tier, core.Component, sqsName))
+                    sqsName)
             fifoQueue=fifoQueue
         /]
 
--- a/aws/components/sqs/setup.ftl
+++ b/aws/components/sqs/setup.ftl
@@ -71,6 +71,10 @@
                 name=dlqName
                 retention=1209600
                 receiveWait=20
+                tags=
+                    getOccurrenceCoreTags(
+                        occurrence,
+                        formatComponentFullName(core.Tier, core.Component, dlqName))
                 fifoQueue=fifoQueue
             /]
         [/#if]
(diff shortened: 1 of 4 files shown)


aws/components/sqs/setup.ftl
```diff
@@ -71,6 +71,10 @@
                 name=dlqName
                 retention=1209600
                 receiveWait=20
+                tags=
+                    getOccurrenceCoreTags(
+                        occurrence,
+                        formatComponentFullName(core.Tier, core.Component, dlqName))
```
roleyfoley, line 77: Might as well keep the name the same as the one used above to name the resource

```suggestion
                        dlqName
```

aws/components/sqs/setup.ftl
```diff
@@ -89,6 +93,10 @@
                 solution.DeadLetterQueue.MaxReceives,
                 solution.DeadLetterQueue.MaxReceives > 0,
                 (environmentObject.Operations.DeadLetterQueue.MaxReceives)!3)
+            tags=
+                getOccurrenceCoreTags(
+                    occurrence,
+                    formatComponentFullName(core.Tier, core.Component, sqsName))
```
roleyfoley, line 99:
```suggestion
                    sqsName
```

Same as above I think those two values end up being the same

roleyfoley (changes requested):

aws/components/sqs/setup.ftl
ken9ross: Cool, done

aws/components/sqs/setup.ftl
ken9ross: Cool, done

awstest/modules/sqs/module.ftl
```diff
@@ -0,0 +1,73 @@
...
+                        "sqsbase" : {
+                            "sqs" : {
+                                "Profiles" : {
+                                    "Testing" : [ "sqsbase" ]
+                                },
+                                "deployment:Unit" : "aws-sqs-base",
+                                "Instances" : {
+                                    "default" : {}
+                                },
+                                "Ordering" : "FirstInFirstOut"
```
roleyfoley, line 27: I don't think that is required as a base level option. FIFO is an optional service

roleyfoley (changes requested):

ml019:
@ken9ross suggest prepending "Fixes " to the issue reference in the PR description - that way the issue is closed when the PR is merged. Its a github feature. There is a range of keywords you can use - ask Dr Google if curious.

ken9ross:
@ml019 I wasn't sure if I should be using "fixes" as this is only in the aws component and not azure (and the initial ticket was in engine - which I'm interpreting as being for all supported engines ie aws and azure). Should I be using "fixes" in this case? ie Are we happy to close an issue if it's fixed for aws and not azure?

awstest/modules/sqs/module.ftl
ken9ross: Done

roleyfoley (approved):

PR #368, opened 2021-08-02
Add tags to ecs cluster services and tasks

ken9ross:
## Intent of Change

## Description
Include tags for ECS Cluster, Service and Tasks

## Motivation and Context
Fixes https://github.com/hamlet-io/engine/issues/1132

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/ecs/state.ftl
+++ b/aws/components/ecs/state.ftl
@@ -226,8 +226,10 @@
     [/#if]
 
     [#local serviceId = formatResourceId(AWS_ECS_SERVICE_RESOURCE_TYPE, core.Id)]
+    [#local serviceName = core.FullName]
     [#local taskId = formatResourceId(AWS_ECS_TASK_RESOURCE_TYPE, core.Id) ]
-    [#local taskName = core.Name]
+    [#local taskName = core.FullName]
+    [#-- taskName changed from core.Name --]
 
     [#local lgId = formatDependentLogGroupId(taskId) ]
     [#local lgName = core.FullAbsolutePath ]
@@ -284,6 +286,7 @@
             "Resources" : {
                 "service" : {
                     "Id" : serviceId,
+                    "Name" : serviceName,
                     "Type" : AWS_ECS_SERVICE_RESOURCE_TYPE,
                     "Monitored" : true
                 },
@@ -366,7 +369,8 @@
     [#local ecsId = parentResources["cluster"].Id ]
 
     [#local taskId = formatResourceId(AWS_ECS_TASK_RESOURCE_TYPE, core.Id) ]
-    [#local taskName = core.Name]
+    [#local taskName = core.FullName]
+    [#-- taskName changed from core.Name --]
     [#local taskRoleId = formatDependentRoleId(taskId)]
 
     [#local executionRoleId = formatDependentRoleId(taskId, "execution")]
--- a/aws/components/ecs/state.ftl
+++ b/aws/components/ecs/state.ftl
@@ -228,8 +228,7 @@
     [#local serviceId = formatResourceId(AWS_ECS_SERVICE_RESOURCE_TYPE, core.Id)]
     [#local serviceName = core.FullName]
     [#local taskId = formatResourceId(AWS_ECS_TASK_RESOURCE_TYPE, core.Id) ]
-    [#local taskName = core.FullName]
-    [#-- taskName changed from core.Name --]
+    [#local taskName = core.Name]
 
     [#local lgId = formatDependentLogGroupId(taskId) ]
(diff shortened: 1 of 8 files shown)


ken9ross:
OK, now I'm confused because it looks like the tests worked when run here, but failed locally

aws/components/ecs/state.ftl
```diff
@@ -226,8 +226,10 @@
     [/#if]
 
     [#local serviceId = formatResourceId(AWS_ECS_SERVICE_RESOURCE_TYPE, core.Id)]
+    [#local serviceName = core.FullName]
     [#local taskId = formatResourceId(AWS_ECS_TASK_RESOURCE_TYPE, core.Id) ]
-    [#local taskName = core.Name]
+    [#local taskName = core.FullName]
+    [#-- taskName changed from core.Name --]
```
roleyfoley, line 232: Where is the task name used? This could cause breaking changes with existing deployments. Renaming resources can be dangerous

aws/components/ecs/state.ftl
```diff
@@ -366,7 +369,8 @@
     [#local ecsId = parentResources["cluster"].Id ]
 
     [#local taskId = formatResourceId(AWS_ECS_TASK_RESOURCE_TYPE, core.Id) ]
-    [#local taskName = core.Name]
+    [#local taskName = core.FullName]
```
roleyfoley, line 372: Where is the task name used? This could cause breaking changes with existing deployments. Renaming resources can be dangerous

awstest/modules/task/module.ftl
```diff
@@ -0,0 +1,127 @@
...
+                    }
+                }
+            },
+            "TestCases" : {
+                "baseecs" : {
+                    "OutputSuffix" : "template.json",
+                    "Tools" : {
+                       "CFNLint" : true
+                    }
+                },
```
roleyfoley, line 81: This isn't really required as we the host is only a underlying dependency. Do we need to perform tests on the solution component when this module is based tasks rather than the host?

awstest/modules/task/module.ftl
```diff
@@ -0,0 +1,127 @@
+[#ftl]
+
+[@addModule
+    name="task"
+    description="Testing module for the aws ecs task component"
+    provider=AWSTEST_PROVIDER
+    properties=[]
+/]
+
```
roleyfoley, line 9: This doesn't appear to be testing tasks components and is actually testing the service component based on the solution module.

roleyfoley (changes requested):

awstest/modules/task/module.ftl
ken9ross: Damn C&P error - sorry

awstest/modules/task/module.ftl
ken9ross: This was me trying (unsuccessfully) to ensure that the soln level was isolated from the app level by having a separate testing profile for ECS

aws/components/ecs/state.ftl
ken9ross: In @createECSTask. I figured it probably would be which is why the comment placeholder, though I thought a mismatched name might be worse

roleyfoley:
testing bug fixed in https://github.com/hamlet-io/engine/pull/1744

aws/components/ecs/state.ftl
ken9ross: Populates task.Name, ARN, and familyName. All of which seem to work fine (unless I'm missing something). Biggest impact I've seen is that it "hides" the significant portion of the task name in the AWS console (and you have to wait for the hover to list the full name of the task definition)

aws/components/ecs/state.ftl
roleyfoley: Lets leave this change out and if you feel it should be investigated lets raise a new ticket  it for it. We need to see what this does with some live deployments.

aws/components/ecs/state.ftl
roleyfoley: Lets leave this out and move it to a separate ticket if it should be changed. It will need to be tested across existing deployments.

roleyfoley (changes requested):

aws/components/ecs/state.ftl
ken9ross: Cool, done

aws/components/ecs/state.ftl
ken9ross: Cool, done

roleyfoley (approved):

PR #371, opened 2021-08-09
Add tagging support for APIgateway

ken9ross:
## Intent of Change

## Description
Add tags to AWS::ApiGateway:: children

## Motivation and Context
Fixes https://github.com/hamlet-io/engine/issues/1130

## How Has This Been Tested?
Locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/state.ftl
+++ b/aws/components/apigateway/state.ftl
@@ -365,6 +365,7 @@ created in either case.
                 },
                 "apideploy" : {
                     "Id" : formatResourceId(AWS_APIGATEWAY_DEPLOY_RESOURCE_TYPE, core.Id, getCLORunId()),
+                    "Name": apiName,
                     "Type" : AWS_APIGATEWAY_DEPLOY_RESOURCE_TYPE
                 },
                 "apistage" : {
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -26,6 +26,7 @@
 
     [#-- Use runId to ensure deploy happens every time --]
     [#local deployId   = resources["apideploy"].Id]
+    [#local deployName = resources["apideploy"].Name]
     [#local stageId    = resources["apistage"].Id]
     [#local stageName  = resources["apistage"].Name]
     [#local stageDependencies = [deployId]]
@@ -464,6 +465,10 @@
                     true
                 )
             outputs=APIGATEWAY_OUTPUT_MAPPINGS
+            tags=
+                getOccurrenceCoreTags(
+                    occurrence,
+                    apiName)
         /]
 
         [@cfResource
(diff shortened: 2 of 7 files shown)


ken9ross:
@roleyfoley I need some guidance. I believe to create an AWS::ApiGateway::DomainName template, I need to add a certificate.

Have I missed something? If not, is there a mechanism for emulating a certificate for testing?

roleyfoley:
The Mocking should return the certificate fake ARN. So just add 


```json 
{ 
   "Certificate" : {} 
}
```

To the configuration of the API inline with how Certificates are configured in the API Gateway schema.

ken9ross:
@roleyfoley I'm getting nowhere in trying to establish a custom domain to test the tagging of AWS::ApiGateway::DomainName. Can I beg some assistance?

roleyfoley:
@ken9ross Sorry missed one extra step, the mapping configuration needs to be added so that it sets up the configuration. Just added it and seems to be create the domain name now

aws/components/user/state.ftl
```diff
@@ -59,7 +59,7 @@
                 },
                 "apikey" : {
                     "Id" : formatDependentResourceId(AWS_APIGATEWAY_APIKEY_RESOURCE_TYPE, userId),
-                    "Name" : core.FullName,
```
roleyfoley, line 62: Names shouldn't have types, the context of the resource already says that

aws/components/apigateway/state.ftl
```diff
@@ -365,6 +365,7 @@ created in either case.
                 },
                 "apideploy" : {
                     "Id" : formatResourceId(AWS_APIGATEWAY_DEPLOY_RESOURCE_TYPE, core.Id, getCLORunId()),
+                    "Name": formatName(apiName,"deploy"),
```
roleyfoley, line 368: Names don't need to include the type. The actual resource sets that context anyway

roleyfoley (changes requested):

aws/components/apigateway/state.ftl
ken9ross: I was assuming the names needed to be unique for the AWS component, so if one hamlet component spawned multiple AWS components, we needed to differentiate between them. 

Do I have that backwards and it should be the hamlet component name (or something else entirely)?

roleyfoley: Ids and Names are generated to be unique as they are formed on the occurrence, so as you split out new occurences the names are always going to be unique across a component. 

Names within AWS are usually for tags which are just metadata and don't need to be unique. If they do need to be unique they only need to be unique within the particular resource .

If you used the same resource across different components then the names would still be unique because all occurrences are unique

ken9ross: Changed, rebased and commit message changed to conventional format

roleyfoley (approved):

PR #375, opened 2021-08-21
refactor: use paramters for az - cfnlint

roleyfoley:
## Intent of Change

## Description

adds a set of AZ lookup and parameter reference functions which can be used to create the parameters with default values and to reference them in the template that they are being used in

## Motivation and Context

cfnlint doesn't recommended hard coding AZ values in templates. While we use generated templates we should try and comply with what cfn lint says and to allow for other cfn lint tests to pass

## How Has This Been Tested?

Tested on local template generation and with test suite

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/cache/setup.ftl
+++ b/aws/components/cache/setup.ftl
@@ -201,12 +201,12 @@
                     multiAZ?then(
                         {
                             "AZMode": "cross-az",
-                            "PreferredAvailabilityZones" : awsZones,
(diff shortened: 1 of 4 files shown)


ml019 (approved):
Seems fundamentally driven by the idea of people manually writing templates. Hamlet doesn't fit the mould.

Issue #254, opened 2021-03-19
[BUG] WAF logging script failing on delete


roleyfoley:
## Current Behaviour

When deleting a loadbalancer or apigateway which uses a script to enable and disable WAF logging an error is reported in the bash script 

```
(Info) Delete completed for devnet-services-channel-app-sharedchannel-api-c4
(Info) Processing epilogue script ...
Note: AWS CLI version 2, the latest major version of the AWS CLI, is now stable and recommended for general use. For more information, see the AWS CLI version 2 installation instructions at: https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2.html

usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help
aws: error: argument --web-acl-id: expected one argument
(Info) Checking WAF Logging state...
```

## Expected Behaviour

The script shouldn't fail and should do whats required to delete

## Possible Solution

Either remove running the waf logging script on delete or handle the cloudformation stack lookup that seems to be failing 

## Steps to Reproduce

1. Run a stop deployment on a deployed load balancer or apigateway with WAF logging enabled

PR #377, opened 2021-08-23
fix(lb): handle missing fqdn

ken9ross:
## Intent of Change

## Description
Handle missing FQDN for LB

## Motivation and Context
Encountered this issue when testing for hamlet-io/engine-plugin-aws#254

## How Has This Been Tested?
Local generation and deployment to AWS

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/lb/setup.ftl
+++ b/aws/components/lb/setup.ftl
@@ -297,7 +297,9 @@
 
                 [/#list]
 
-                [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [#if fqdn?has_content]
+                    [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [/#if]
             [/#if]
 
             [#-- Redirect rule processing --]
--- a/aws/components/lb/setup.ftl
+++ b/aws/components/lb/setup.ftl
@@ -297,10 +297,9 @@
 
                 [/#list]
 
-                [#if fqdn?has_content]
-                    [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
-                [#else]
-                    [@fatal message="Missing FQDN" context=solution /]
+                [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [#if ! fqdn?has_content]
+                    [@fatal message="HostName/Certificate property is required when HostFilter:true " context=solution /]
                 [/#if]
             [/#if]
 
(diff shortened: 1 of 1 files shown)


ken9ross:
@roleyfoley I think this is all that is required to rectify this

aws/components/lb/setup.ftl
```diff
@@ -297,7 +297,9 @@
 
                 [/#list]
 
-                [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [#if fqdn?has_content]
+                    [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [/#if]
```
roleyfoley, line 302: I think this need to be an exception rather than a bypass. The host filter option is more a security service so needs to be explicit rather still allow it to happen.

roleyfoley (changes requested):

aws/components/lb/setup.ftl
ken9ross: So a warning message (but still allow it through), or a fatal if no domainRedirectRules were found, or something else?

roleyfoley: If hostFilter is enabled and there is no fqdn then it should be a fatal. We can't do what the user has requested

ken9ross: Cool, done

aws/components/lb/setup.ftl
```diff
@@ -299,6 +299,8 @@
 
                 [#if fqdn?has_content]
                     [#local listenerRuleConditions += getListenerRuleHostCondition(fqdn) ]
+                [#else]
+                    [@fatal message="Missing FQDN" context=solution /]
```
roleyfoley, line 303: I think this needs to be more descriptive since we don't have a one to one map between the FQDN and a solution configuration options.

roleyfoley: We also don't want the error to hide the template content so the fatal should be in its own if statement so that we can still see the template content when debugging

roleyfoley (changes requested):

aws/components/lb/setup.ftl
ken9ross: For the first comment, I thought the context would carry that information, have I missed something (the message was just what was wrong)

For the second, cool will do

roleyfoley: If that error comes up can you figure out how to set the FQDN?

ken9ross: No, but I was assuming that was my ignorance as it comes from `resources["listenerRule"].FQDN` where resources is the same subOccurrence as the context in the error message

roleyfoley (approved):

PR #386, opened 2021-08-30
feat(iam): tags

ken9ross:
## Intent of Change

## Description
Add tags for IAM::Users and IAM::Roles

## Motivation and Context
Fixes https://github.com/hamlet-io/engine/issues/1129

## How Has This Been Tested?
Local generation and extend test cases for users

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/services/iam/resource.ftl
+++ b/aws/services/iam/resource.ftl
@@ -106,6 +106,7 @@
         ]
     [/#list]
 
+    [#local constructedName=""]
     [#if ! name?has_content]
         [#local constructedName=name]
     [#else]
--- a/aws/services/iam/resource.ftl
+++ b/aws/services/iam/resource.ftl
@@ -95,6 +95,7 @@
             managedArns=[]
             policies=[]
             dependencies=[] 
+            tags=[]
             ]
 
     [#local trustedAccountArns = [] ]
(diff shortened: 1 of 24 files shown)


aws/services/iam/resource.ftl
```diff
@@ -105,6 +106,12 @@
         ]
     [/#list]
 
+    [#local constructedName=""]
+    [#if ! name?has_content]
+        [#local constructedName=name]
+    [/#if]
```
roleyfoley, line 112: Do we have the constructed name anymore? if not I think this can be removed and a check for a name parameter would be used?

roleyfoley (changes requested):

aws/services/iam/resource.ftl
ken9ross: True, done

aws/services/iam/resource.ftl
```diff
@@ -105,6 +106,8 @@
         ]
     [/#list]
 
+    [#local tags=getOccurrenceCoreTags(occurrence, name) ]
+
```
roleyfoley, line 110: The occurrence isn't passed in as a parameter here. How would the occurrence be found?

roleyfoley (changes requested):

aws/services/iam/resource.ftl
ken9ross: You're right. I'm now very confused as to why no error, and where the value came from. That's at odds with the freemarker doco on local variables (incl parameters) unless I'm missing something

ken9ross: Updated to pass tags and handle legacy account not passing anything to use getCfTemplateCoreTags(name)

roleyfoley (approved):

PR #391, opened 2021-09-08
Feat aws pinpoint

kshychko:
## Intent of Change

## Description

This PR adds an initial support for AWS Pinpoint component creation as a service for the shared correspondent component.
## Motivation and Context

To support push notifications for mobile applications AWS Pinpoint project needs to be created and its projectID shared with mobile apps and notifications services providers. Also, Amplify requires Cognito setup with identity roles providing the access to AWS Pinpoint project.
## How Has This Been Tested?

Tested locally, on development deployment and with test cases
### Prerequisite PRs:

- hamlet-io/engine#1793

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/firewall/setup.ftl
+++ b/aws/components/firewall/setup.ftl
@@ -49,13 +49,6 @@
         tags=getOccurrenceCoreTags(occurrence, core.FullName)
     /]
 
-                            {
-                                "Fn::Join": [
-                                    ",",
-                                    tierSubnetIdRefs
-                                ]
-                            },
-
     [@cfOutput
         formatId(firewallId, INTERFACE_ATTRIBUTE_TYPE),
         {
--- a/aws/components/correspondent/id.ftl
+++ b/aws/components/correspondent/id.ftl
@@ -1,0 +1,12 @@
+[#ftl]
+
+[@addResourceGroupInformation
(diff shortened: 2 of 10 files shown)


kshychko:
@roleyfoley, could you please check the test case configuration. It seems correct to me, but I was getting a "test case not found" error when was running the test locally.

aws/components/firewall/setup.ftl
```diff
@@ -49,13 +49,6 @@
         tags=getOccurrenceCoreTags(occurrence, core.FullName)
     /]
 
-                            {
-                                "Fn::Join": [
-                                    ",",
-                                    tierSubnetIdRefs
-                                ]
-                            },
-
```
roleyfoley, line 58: Not sure whats happened with this? might need the rebase?

roleyfoley (changes requested):

aws/components/firewall/setup.ftl
kshychko: @roleyfoley, I removed it as noticed this unused code while was using the firewall component as a reference.

roleyfoley: Oh sorry its just floating in the middle of nowhere! Thanks for catching that

roleyfoley (approved):

PR #392, opened 2021-09-09
feat(ds): New Component - Directory Services

ken9ross:
## Intent of Change

## Description

Add new component "directory" and service "ds"

**Note**: I have deliberately not included the createAlias feature as this has warning associated with it that it can not be reused (ie http://_alias_.awsapps.com is permanently unique)

## Motivation and Context

Fixes https://github.com/hamlet-io/engine/issues/1756

## How Has This Been Tested?

Locally generated, new test cases and deployed to AWS. Note deployment is very slow (30 minutes)

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1796

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/directory/setup.ftl
+++ b/aws/components/directory/setup.ftl
@@ -11,10 +11,6 @@
     [#local solution = occurrence.Configuration.Solution ]
     [#local resources = occurrence.State.Resources ]
 
-    [#local core = occurrence.Core ]
-    [#local solution = occurrence.Configuration.Solution ]
-    [#local resources = occurrence.State.Resources ]
-
     [#-- Baseline component lookup --]
     [#local baselineLinks = getBaselineLinks(occurrence, [ "Encryption" ] )]
     [#local baselineComponentIds = getBaselineComponentIds(baselineLinks)]
@@ -31,32 +27,38 @@
     [#local networkResources = networkLinkTarget.State.Resources ]
     [#local vpcId = networkResources["vpc"].Id ]
 
+    [#local dnsPorts = [ 
+        "dns-tcp", "dns-tcp",
+        "globalcatalog",
+        "kerebosauth88-tcp", "kerebosauth88-udp", "kerebosauth464-tcp", "kerebosauth464-udp",
+        "ldap-tcp", "ldap-udp", "ldaps",
+        "netlogin-tcp", "netlogin-udp", 
+        "ntp",
+        "rpc", "ephemeralrpctcp", "ephemeralrpcudp",
+        "rsync",
+        "smb-tcp", "smb-udp", 
+        "anyicmp"
+    ]]
+
     [#-- Resources and base configuration --]
     [#local dsId = resources["directory"].Id ]
-    [#local dsFullName = resources["directory"].Name ]
+    [#local fqdName = resources["directory"].Name ]
     [#local dsShortName = resources["directory"].ShortName ]
     [#local securityGroupId = resources["sg"].Id ]
     [#local securityGroupName = resources["sg"].Name ]
 
     [#local engine = solution.Engine]
+    [#local enableSSO = solution["aws:EnableSSO"]]
 
     [#switch engine]
-        [#case "aws:ms-std"]
+        [#case "ActiveDirectory"]
             [#local type = "MicrosoftAD"]
-            [#local edition = "Standard"]
-            [#local size = ""]
+            [#local size = (solution.Size == "Standard")?then("Standard","Enterprise")]
             [#break]
 
-        [#case "aws:ms-ent"]
-            [#local type = "MicrosoftAD"]
-            [#local edition = "Enterprise"]
-            [#local size = ""]
-            [#break]
-
-        [#case "simple"]
-            [#local type = "Simple"]
-            [#local edition = ""]
-            [#local size = (solution.Size!"Small" = "Small")?then("Small","Large")]
+        [#case "Simple"]
+            [#local type = "SimpleAD"]
+            [#local size = (solution.Size == "Small")?then("Small","Large")]
             [#break]
 
         [#default]
@@ -66,7 +68,7 @@
                 detail="Unsupported engine provided"
             /]
             [#local type = "unknown" ]
-            [#local edition = "unknown" ]
+            [#local size = "unknown" ]
 
             [#break]
     [/#switch]
@@ -94,9 +96,10 @@
             [#if deploymentSubsetRequired(DIRECTORY_COMPONENT_TYPE, true)]
                 [@createSecurityGroupRulesFromLink
                     occurrence=occurrence
-                    groupId=cacheSecurityGroupId
+                    groupId=securityGroupId
                     linkTarget=linkTarget
-                    inboundPorts=[ "ssh" ]
+                    inboundPorts=dnsPorts
+                    networkProfile=networkProfile
                 /]
             [/#if]
 
@@ -108,8 +111,7 @@
     [#local passwordSecretKey = "password" ]
 
     [#if secretStoreLink?has_content ]
-
-        [@setupComponentSecret
+       [@setupComponentSecret
             occurrence=occurrence
             secretStoreLink=secretStoreLink
             kmsKeyId=cmkKeyId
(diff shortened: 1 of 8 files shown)


aws/components/directory/setup.ftl
```diff
@@ -0,0 +1,254 @@
...
+            groupId=securityGroupId
+            networkRule=ingressNetworkRule
+        /]
+
+        [#if !hibernate]
+
+            [#if ! testMaintenanceWindow(solution.MaintenanceWindow)]
+                [@fatal message="Maintenance window incorrectly configured" context=solution /]
+                [#return]
+            [/#if]
```
roleyfoley, line 168: Does the AD service have maintenance windows?

roleyfoley (changes requested):

aws/components/directory/setup.ftl
```diff
@@ -0,0 +1,254 @@
...
+            [#local type = "MicrosoftAD"]
+            [#local edition = "Enterprise"]
+            [#local size = ""]
+            [#break]
+
+        [#case "simple"]
+            [#local type = "Simple"]
+            [#local edition = ""]
+            [#local size = (solution.Size!"Small" = "Small")?then("Small","Large")]
+            [#break]
```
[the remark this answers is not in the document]
ken9ross: Yes it is. I was expecting Standard to be a cutdown, but it's not. Done

aws/components/directory/setup.ftl
ken9ross: No, Missed it why slicing and dicing

aws/components/directory/setup.ftl
```diff
@@ -0,0 +1,254 @@
...
+            name=securityGroupName
+            vpcId=vpcId
+            occurrence=occurrence
+        /]
+
+        [@createSecurityGroupRulesFromNetworkProfile
+            occurrence=occurrence
+            groupId=securityGroupId
+            networkProfile=networkProfile
+            inboundPorts=brokerPorts
```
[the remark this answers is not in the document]
ken9ross: Correct

aws/components/directory/state.ftl
```diff
@@ -0,0 +1,61 @@
...
+            "Roles" : {
+                "Inbound" : {
+                    "networkacl" : {
+                        "SecurityGroups" : securityGroupId,
+                        "Description" : core.FullName
+                    }
+                },
+                "Outbound" : {
+                    "networkacl" : {
+                        "Ports" : [ "ssh" ],
```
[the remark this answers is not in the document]
ken9ross: Yes it is. I was just using a placeholder. Done

aws/components/directory/setup.ftl
```diff
@@ -0,0 +1,221 @@
...
+    [#local securityGroupId = resources["sg"].Id ]
+    [#local securityGroupName = resources["sg"].Name ]
+
+    [#local engine = solution.Engine]
+    [#local enableSSO = solution["aws:EnableSSO"]]
+
+    [#switch engine]
+        [#case "ActiveDirectory"]
+            [#local type = "MicrosoftAD"]
+            [#local size = (solution.Size == "Standard")?then("Standard","Enterprise")]
```
roleyfoley, line 56: This should probably be handled on the component configuration rather than within the setup. 

I can see they are different between the different implementations so we can either make our own values to align with it or specific values.

So on the core provider we either have values as small medium large, and then translate them in the backend or make sizes vendor specific.

aws/components/directory/setup.ftl
```diff
@@ -0,0 +1,221 @@
...
+                            "Fn::GetAtt": [
+                                dsId,
+                                "Alias"
+                            ]
+                        }
+                    ]
+                },
+                false
+            /]
+--]
```
roleyfoley, line 216: Do we need this? What would the IP addresses be used for? The dhcp scopes? and we can probably remove the commented out one or add comments why its like that

aws/components/directory/state.ftl
```diff
@@ -0,0 +1,80 @@
...
+        "globalcatalog",
+        "kerebosauth88-tcp", "kerebosauth88-udp", "kerebosauth464-tcp", "kerebosauth464-udp",
+        "ldap-tcp", "ldap-udp", "ldaps",
+        "netlogin-tcp", "netlogin-udp", 
+        "ntp",
+        "rpc", "ephemeralrpctcp", "ephemeralrpcudp",
+        "rsync",
+        "smb-tcp", "smb-udp", 
+        "anyicmp"
+    ]]
```
roleyfoley, line 29: To remove duplication of quite a bit list,  you could put these as an attribute on the directory resource as Ports or something like that. Then in the setup routine you can just reference the resource details instead of having to duplicate the array. 
One niggly thing, these are only the ports when using AD not when you are using simple AD right? Do they both support stuff like RPC?

aws/services/ds/resource.ftl
```diff
@@ -0,0 +1,50 @@
...
+]
+[@addOutputMapping
+    provider=AWS_PROVIDER
+    resourceType=AWS_DIRECTORY_RESOURCE_TYPE
+    mappings=DIRECTORY_OUTPUT_MAPPINGS
+/]
+
+[#macro createDSInstance id name
+    type
+    size
```
roleyfoley, line 18: Probably good to set the size with a default of "" since its not going to be required for everything right?

awstest/modules/directory/module.ftl
```diff
@@ -0,0 +1,220 @@
...
+            "TestCases" : {
+                "directorysimplebase" : {
+                    "OutputSuffix" : "template.json",
+                    "Tools" : {
+                       "CFNLint" : true
+                    },
+                    "Structural" : {
+                        "CFN" : {
+                            "Resource" : {
+                                "queueHost" : {
```
roleyfoley, line 84: Copy paste

awstest/modules/directory/module.ftl
```diff
@@ -0,0 +1,220 @@
...
+            "TestCases" : {
+                "directoryadbase" : {
+                    "OutputSuffix" : "template.json",
+                    "Tools" : {
+                       "CFNLint" : true
+                    },
+                    "Structural" : {
+                        "CFN" : {
+                            "Resource" : {
+                                "queueHost" : {
```
roleyfoley, line 188: CopyPaste

roleyfoley (changes requested):

aws/components/directory/setup.ftl
ken9ross: Made it Small and Large from the feedback on engine-core

aws/components/directory/setup.ftl
ken9ross: The IP addresses of the DNS servers for the directory, I was thinking about the possibility of pulling these on a windows server being created with a link to the DS

The commented out block has to do with the note in the description about my deliberately not included the createAlias feature. Left it there in case you wanted to say it was needed

aws/services/ds/resource.ftl
ken9ross: When I collapsed edition and size, it becomes mandatory (or at least defaulted to Small or Large)

aws/components/directory/state.ftl
ken9ross: You're correct RPC is not, I've left LDAPS in (even though it is recommended against) and RSYNC as I'm not sure on the replication used for SimpleAD

awstest/modules/directory/module.ftl
ken9ross: Done

awstest/modules/directory/module.ftl
ken9ross: Done

aws/components/directory/setup.ftl
roleyfoley: Capturing the output won't change if its defined or not, just that it would be saved in our local outputs. Are you saying you haven't implemented the alis configuration?

ken9ross: OK, so we could still access it from the console.

Correct, I haven't implemented the alias configuration. Should I?

roleyfoley (approved):

PR #394, opened 2021-09-09
refactor(lb): support for default rule control

roleyfoley:
## Intent of Change

## Description

- Adds support for configuring the default action on a listener - this aligns with any of the standard rules actions available for the listener rules ( forward, fixed, redirect etc ) 
- Removes the default target group that was included to act as a dummy endpoint and instead sends a fixed response by default 

## Motivation and Context

- Removes the creation of a target group that will never be used within hamlet 
- Allows for integration with template deployments which create their own listener rules and rely on the default for the base actions 

## How Has This Been Tested?

Tested locally and on a development deployment 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/lb/setup.ftl
+++ b/aws/components/lb/setup.ftl
@@ -280,6 +280,24 @@
                         [#local path = solution.Path ]
                     [/#if]
                 [/#if]
+
+
+                [#if listenerRulePriority?is_string &&
+                            listenerRulePriority == "default" ]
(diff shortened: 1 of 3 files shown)


ml019 (approved):

ml019:
@roleyfoley  thought the test failures might have been because the engine PR wasn't merged but looks like something in the test suite still isn't happy.

ml019 (approved):

PR #421, opened 2021-09-29
refactor: setContext wrapper functions (1)

ml019:
## Intent of Change

## Description
Replace the use of global variables from `setContext()` with accessor functions for the same information.

## Motivation and Context
This is in preparation for full district support, where all config and state access needs to be via the input state.

### Prerequisite PRs:
- https://github.com/hamlet-io/engine/pull/1821

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
Tests will fail until engine PR is merged.

ml019:
@roleyfoley I think we need the production version of the hamlet cli that has the updated linting support? The tests have been updated by the checking code is expecting the old config.

roleyfoley (approved):

PR #422, opened 2021-09-30
feat(awswin): Add app whitelisting (incomplete)

ken9ross:
## Intent of Change

## Description

Add App Whitelisting computetask for EMA (only generates script does not execute the script)

## How Has This Been Tested?

Locally generated and deployed to AWS

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1822

### Dependent PRs:

- None

### Consumer Actions:

- None


ken9ross:
@roleyfoley These are the work I've done on AppLocker. Don't know if they're any use with the migration to SSM or not, but just in case

roleyfoley:
Thanks @ken9ross if this isn't ready to merge could you please mark this as a Draft PR
You can do that under the reviewers section

ken9ross:
@roleyfoley It's safe to merge, but it only creates scripts to do the work, it doesn't run them (so it doesn't solve any business problem). I was imagining you would either merge to be a start point or delete as irrelevant depending on whether they were useful for the SSM alternative you suggested.

Regardless, done

PR #428, opened 2021-10-13
fix(ds): Remove redundant sg and modify sg to align with IPAddressGroups

ken9ross:
## Intent of Change

## Description
Remove code which created irrelevant security group
Epilogue script change to modify AWS created security group to align with defined IPAddressGroups

## Motivation and Context

Fixes https://github.com/hamlet-io/engine/issues/1832

## How Has This Been Tested?

Locally generated and deployed to AWS

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/directory/setup.ftl
+++ b/aws/components/directory/setup.ftl
@@ -43,34 +43,11 @@
         [#case "ActiveDirectory"]
             [#local type = "MicrosoftAD"]
             [#local size = (solution.Size == "Small")?then("Standard","Enterprise")]
-            [#local dnsPorts = [
-                "dns-tcp", "dns-udp",
-                "globalcatalog",
-                "kerebosauth88-tcp", "kerebosauth88-udp", "kerebosauth464-tcp", "kerebosauth464-udp",
-                "ldap-tcp", "ldap-udp", "ldaps",
-                "netlogin-tcp", "netlogin-udp",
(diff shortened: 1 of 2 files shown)


aws/components/directory/setup.ftl
```diff
@@ -257,4 +152,54 @@
         [/#if]
     [/#if]
 
+    [#local epilogue_cidr=[]]
+    [#local epilogue_content=[]]
+
+    [#if ! solution.IPAddressGroups?seq_contains("_global")]
+        [#local epilogue_content += [
+                "   info \"SecGroupId = $\{secgrp_id}\"",
```
roleyfoley, line 160: You can use the r' syntax which is the equivalent of literal strings 

For example 
https://github.com/hamlet-io/engine-plugin-aws/blob/3c1b1337fb55f8b06d2e7846da7d9936f5986999/aws/extensions/computetask_awslinux_eip/extension.ftl

roleyfoley (changes requested):

aws/components/directory/setup.ftl
ken9ross: Not sure I agree it adds to the readability as the complex lines have quotes within quotes, so there's no avoiding escaping values, but done

PR #456, opened 2021-11-11
fix: Kinesis delivery stream S3 permissions

ml019:
## Intent of Change
- Breaking change (fix or feature that would cause existing functionality to change)

## Description
- when a delivery stream destination is an S3 bucket, ensure that the encryption configuration aligns with that of the bucket
- name processing log streams to better reflect their function
- correct an issue where creation of log subscriptions in an lg subset was failing because the iam resources needed to access a delivery stream were not being created.

## How Has This Been Tested?
- Template processing was failing
- Log consolidation was failing due to the presence of KMS encrypted objects

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/datafeed/state.ftl
+++ b/aws/components/datafeed/state.ftl
@@ -38,13 +38,13 @@
                     },
                     "backuplgstream" : {
                         "Id" : formatDependentResourceId(AWS_CLOUDWATCH_LOG_GROUP_STREAM_RESOURCE_TYPE, lgId, "backup"),
-                        "Name" : "S3Delivery",
+                        "Name" : "Backup",
                         "Type" : AWS_CLOUDWATCH_LOG_GROUP_STREAM_RESOURCE_TYPE,
                         "IncludeInDeploymentState" : false
                     },
                     "streamlgstream" : {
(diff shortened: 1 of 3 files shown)


aws/components/datafeed/state.ftl
```diff
@@ -38,13 +38,13 @@
                     },
                     "backuplgstream" : {
                         "Id" : formatDependentResourceId(AWS_CLOUDWATCH_LOG_GROUP_STREAM_RESOURCE_TYPE, lgId, "backup"),
-                        "Name" : "S3Delivery",
+                        "Name" : "Backup",
```
roleyfoley, line 41: I think the names listed are the ones used by Firehose. This name is used on the streams and I don't think they can be changed

roleyfoley (changes requested):

aws/components/datafeed/state.ftl
ml019: OK - I'll revert them...

ml019:
@roleyfoley ready for review again

roleyfoley (approved):

PR #459, opened 2021-11-16
Add support for WAFv2

ken9ross:
## Intent of Change

## Description

Add support for WAFv2 controlled by Version attribute

## Motivation and Context

Fixes https://github.com/hamlet-io/engine/issues/1768

## How Has This Been Tested?

Local generation and deployed to AWS

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1862

### Dependent PRs:

- None

### Consumer Actions:

- None


ken9ross:
I'm sure there has to be a better way to address Arn in the state.ftl files, but I couldn't find it. Any advice?

aws/components/cdn/setup.ftl
```diff
@@ -14,10 +14,10 @@
     [#local cfName              = resources["cf"].Name]
 
     [#local wafPresent          = isPresent(solution.WAF) ]
-    [#local wafVersion          = (solution.WAF.Version)!"V1" ]
+    [#local wafVersion          = (solution.WAF.Version)!"v1" ]
```
ml019, line 17: This is inconsistent with the defaults in the engine which sets v2 as the default. Given it is being defaulted via the engine configuration, do we need to catch a missing value at this point?

I wonder if initially it would be better to leave the default as v1 so existing deployments aren't affected and new ones can pick v2 from the get go. At some future point, we then change the default to v2 once we are confident the bulk of installations have transitioned.

ml019 (changes requested):

ml019:
@ken9ross please review the conflict - probably a result of all the changes I made related to consolidated logging.

aws/components/cdn/setup.ftl
ken9ross: Because solution.WAF may be absent (refer L16), I added that to make sure that it was always a text string even when absent to simplify the code. While I may have made an error in working through it, I did hit issues when WAF was not expected to be set

I had it that way initially for that reason. Happy to revert

ml019 (approved):

PR #462, opened 2021-11-16
Outbound MTA email identity support

ken9ross:
## Intent of Change

## Description

Extend epilogue to include using sesv2 put|create-email-identity to configure default routing

## Motivation and Context

Fixes https://github.com/hamlet-io/engine/issues/1861

## How Has This Been Tested?

Locally generated and deployed to AWS, but email routing not tested

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/mta/setup.ftl
+++ b/aws/components/mta/setup.ftl
@@ -91,16 +91,26 @@
                             [#local delBash = []]
                             [#list solution.Conditions.Recipients as emailAddr]
                                 [#local addBash += [
-                                    r'    info "Create email identity"',
-                                    r'    aws --region "' + getRegion() + r'" sesv2 create-email-identity --email-identity ' + emailAddr + r' --configuration-set-name ' + configSetName
+                                    r'    info "Create email identity configset"',
+                                    r'    if [ -z "$(aws --region "' + getRegion() + r'" ses list-identities | grep "' + emailAddr + r'")" ]; then ',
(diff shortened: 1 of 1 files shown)


ken9ross:
BTW There are also some additional fixes for inboundmta which seemed to still have some of the merging topicArn errors, so please check I haven't broken those too

aws/components/mta/setup.ftl
```diff
@@ -109,12 +137,21 @@
...
+                                ] + updBash + [
+                                    r'    ;;'
+                                ]+
+                                [
+                                    r'  delete)'
+                                ] + delBash + [
+                                    r'    ;;'
+                                ]+
+                                [
+                                    "esac"
```
roleyfoley, line 154: It might be cleaner here to just do this as a separate case statement. It looks like the addBash and updBash are the same for identities. That would also remove these generic variables which don't really say what this is for.

aws/components/mta/setup.ftl
```diff
@@ -85,6 +86,33 @@
                     [#list topicArns as topicArn ]
                         [#-- notifications must be done through CLI --]
                         [#if deploymentSubsetRequired("epilogue", false) ]
+                            [#local addBash = []]
+                            [#local updBash = []]
+                            [#local delBash = []]
```
roleyfoley, line 91: Lets merge recipient details into a single command that is added to the epilogue script. then this can be a singe variable for looking after the recipient assignemnt

ml019 (approved):

PR #476, opened 2021-11-21
feat(apigateway): Execution Log Control

ml019:
## Intent of Change

## Description
Support subscriptions and retention control on the execution log.

## Motivation and Context
The primary driver for this change is to permit the inclusion of execution logs in the log consolidation process via a log subscription.

Because the api gateway does not permit control over the execution log group name and includes the api id in the name it creates, special logic is required to control log group retention (normally set when creating the log group) and to create log subscriptions on the log group, for instance for log consolidation.

The change creates a pseudo stack for the execution log group and sets its retention according to the operational expiration setting.

In order that subscription processing works the same as if the execution log group could be created, the deployment unit of the pseudo stack is set to match what is used for the access log group. This also makes for a straightforward transition in the future to explicit control of the log group creation if/when AWS supports it.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:
- engine - https://github.com/hamlet-io/engine/pull/1869
- executor-bash - https://github.com/hamlet-io/executor-bash/pull/300

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -20,6 +20,10 @@
     [#local apiId      = resources["apigateway"].Id]
     [#local apiName    = resources["apigateway"].Name]
 
+    [#-- Execution Log Group --]
+    [#local executionLgId   = resources["lg"].Id]
+    [#local executionLgName = resources["lg"].Name]
+
     [#-- Access Log Group --]
     [#local accessLgId   = resources["accesslg"].Id]
     [#local accessLgName = resources["accesslg"].Name]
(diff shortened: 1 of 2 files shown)


aws/components/apigateway/setup.ftl
```diff
@@ -473,6 +477,18 @@
         [/#if]
     [/#if]
 
+    [#-- hamlet doesn't create the execution log group, but can manage any subscriptions   --]
+    [#-- Note that subscriptions will only be created on the second and subsequent stack   --]
+    [#-- execution, i.e. when the epilogue script has been successfully run at least once. --]
+    [#if getExistingReference(executionLgId)?has_content]
```
roleyfoley, line 483: Might be good to add an @warn that you need to run a second pass

roleyfoley (changes requested):

aws/components/apigateway/setup.ftl
ml019: Yup - good idea.

roleyfoley (approved):

PR #480, opened 2021-12-02
fix(cloudwatchslack): change topic priority

roleyfoley:
## Intent of Change

## Description

- Sets the deployment priority to ensure the topic is deployed before other solution components which are creating alerts that need the topic
- Updates the component format to align with the latest versions

## Motivation and Context

Reduces the need to redeploy components to enable the alerts as the topic is required by cloudwatch to set the alert action

## How Has This Been Tested?

tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


aws/modules/cloudwatchslack/module.ftl
```diff
@@ -102,48 +102,41 @@
...
+                                            "Tier" : tier,
+                                            "Component" : lambdaName,
+                                            "Function" : "send"
                                         }
                                     }
                                 }
                             }
                         },
                         lambdaName : {
                             "lambda" : {
```
ml019, line 121: Given the explicit use of the `attribute`, I think this level in the config can be removed?

ml019 (changes requested):

aws/modules/cloudwatchslack/module.ftl
roleyfoley: Where do you mean attribute? couldn't see it mentioned

ml019: Line 121 is the type based attribute "lambda", line 122 is the type attribute. I noticed elsewhere you'd dropped the type attribute (e.g. "topic") so for consistency we should drop this one.

roleyfoley: Ah right! sorry missed that will fix them up

ml019 (approved):

PR #483, opened 2021-12-12
feat: ensure inline code changes on update

roleyfoley:
## Intent of Change

## Description

- Adds the ability to include the runId as part of inline code
- Moves the cfredirect extension over to the aws provider
- Adds support for a fixed path when performing redirects
- Sets the deletion policy for lambda versions to retain

## Motivation and Context

The main motivation around these changes is to handle updates to lambda@edge functions that are deployed using inline code.

- Including the run Id in inline zip files ensures that a new version can be created when version on deploy is used
- Moving cfredirect aligns with the extension being for cloudfront/ lambda@edge only
- The fixed path is for redirects between different services where paths might not align
- The deletion policy ensures that existing versions are retained when cloudfront keeps hold of them

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/lambda/setup.ftl
+++ b/aws/components/lambda/setup.ftl
@@ -94,10 +94,10 @@
                     environmentName,
                     segmentName,
                     fn,
-                    solution.Image.UrlSource.Url,
+                    solution.Image["source:Url"].Url,
                     "lambda",
                     "lambda.zip",
-                    solution.Image.UrlSource.ImageHash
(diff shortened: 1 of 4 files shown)


ml019 (approved):

ml019:
@roleyfoley looks like a few attribute set definitions with pluraltype definitions that got missed with the engine PR that removed them...

PR #486, opened 2021-12-16
fix: s3 versioning without lifecycle management

ml019:
## Intent of Change

## Description
Because of the location of the configuration controlling versioning of the s3 bucket, if versioning is turned on, the bucket will automatically be lifecycled.

## Motivation and Context
To permit versioning without lifecycle management, support a separate versioning attribute.

The Versioning attribute on Lifecycle will be deprecated to avoid lifecycle management as an unwanted side effect of enabling versioning.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:
- https://github.com/hamlet-io/engine/pull/1884

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/s3/setup.ftl
+++ b/aws/components/s3/setup.ftl
@@ -19,7 +19,7 @@
 
     [#local roleId = resources["role"].Id ]
 
-    [#local versioningEnabled = solution.Lifecycle.Versioning ]
+    [#local versioningEnabled = solution.Versioning!solution.Lifecycle.Versioning ]
 
     [#local replicationEnabled = false]
     [#local replicationConfiguration = {} ]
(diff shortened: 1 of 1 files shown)


aws/components/s3/setup.ftl
```diff
@@ -19,7 +19,7 @@
 
     [#local roleId = resources["role"].Id ]
 
-    [#local versioningEnabled = solution.Lifecycle.Versioning ]
+    [#local versioningEnabled = solution.Versioning!solution.Lifecycle.Versioning ]
```
roleyfoley, line 22: Lets add a @warn here with the deprecation details if its found in the lifecycle section as configured.

roleyfoley (changes requested):

aws/components/s3/setup.ftl
ml019: Done

ml019:
@roleyfoley ready for another review

roleyfoley (approved):

PR #487, opened 2021-12-18
fix(apigateway): execution log lookup and waf Id

roleyfoley:
## Intent of Change

## Description

- Gets the api id from the cloudformation stack so that it can be found on create actions
- fixes an issue where the wrong id lookup was used on v2 waf lookups

## Motivation and Context

Deployments failing on initial setup as the log group doesn't exist until the api is invoked 

## How Has This Been Tested?

tested locally

## Related Changes

- https://github.com/hamlet-io/executor-bash/pull/302

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -1186,6 +1186,14 @@
             [/#if]
         [/#list]
 
+
+        [#local lgLookupName = [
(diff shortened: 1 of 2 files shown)


ml019 (approved):
Good catch. I think I tested retention settings on existing stacks but not new ones.

PR #524, opened 2022-02-21
feat: add run ecs task support for runbooks

roleyfoley:
## Intent of Change

## Description

- Adds task to support executing an AWS task with container overrides
- Adds capacity provider state support for tasks
- Adds state for the required network resources to run tasks

## Motivation and Context

Allows for the replacement and user customisation of running on demand container tasks on AWS ECS

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/ecs/id.ftl
+++ b/aws/components/ecs/id.ftl
@@ -157,6 +157,25 @@
         {
             "Names" : "NetworkMode",
             "Values" : [ "shared:awsvpc", "awsvpc" ]
+        },
+        {
+            "Names" : "Placement",
+            "Children" : [
+                {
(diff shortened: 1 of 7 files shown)


ml019:
How does this differ from run ecs command?

roleyfoley:
> How does this differ from run ecs command?

This aligns with the current run task bash script that we use. Essentially non-interactive and also doesn't require the ECS run plugin that uses websockets and a go plugin to be installed.

ml019 (approved):

PR #537, opened 2022-03-29
feat(lb): add support for alb as a network target

roleyfoley:
## Intent of Change

## Description

- Adds support for using an application load balancer as a network load balancer fixed target
- Updates healthchecks to support using HTTP or HTTPS protocol when using TCP as the target protocol

## Motivation and Context

This is an awsism for load balancing and is useful when you need some advanced networking features such as 
- Private Link sharing of applications between VPCS 
- Static IP Addresses for application load balancers ( NLB gets fixed IP's assigned to them ) 

See details on why here https://docs.aws.amazon.com/elasticloadbalancing/latest/network/application-load-balancer-target.html 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/lb/id.ftl
+++ b/aws/components/lb/id.ftl
@@ -28,3 +28,19 @@
             AWS_CERTIFICATE_MANAGER_SERVICE
         ]
 /]
+
+[@addResourceGroupAttributeValues
+    type=LB_PORT_COMPONENT_TYPE
+    provider=AWS_PROVIDER
+    extensions=[
+        {
+            "Names" : "Forward",
+            "Children" : [
+                {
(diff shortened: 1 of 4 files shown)


ml019 (approved):

ml019:
Looks like the definition of `getHealthCheckProtocol()` is missing?

roleyfoley:
> Looks like the definition of `getHealthCheckProtocol()` is missing?

Forgot to mention that this was dependent on https://github.com/hamlet-io/engine/pull/1963

Reran the tests and looks like its all good now

PR #538, opened 2022-03-29
feat(cdn): add error for wrong logging region

roleyfoley:
## Intent of Change

## Description

- Adds a fatal exception to outline that logging can only be enabled if the CDN is deployed to us-east-1. the firehose stream must be in us-east-1

## Motivation and Context

Stops from deployments happening where logs are required but the CDN has already been deployed

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/cdn/setup.ftl
+++ b/aws/components/cdn/setup.ftl
@@ -508,6 +508,17 @@
 
     [#if wafPresent ]
         [#if solution.WAF.Logging.Enabled]
+
+            [#if getRegion() != "us-east-1" ]
+                [@fatal
+                    message="To enable fireshose based logging for WAF on CDN the deployment must be run from us-east-1"
+                    context={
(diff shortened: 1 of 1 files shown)


aws/components/cdn/setup.ftl
```diff
@@ -508,6 +508,17 @@
 
     [#if wafPresent ]
         [#if solution.WAF.Logging.Enabled]
+
+            [#if getRegion() != "us-east-1" ]
+                [@fatal
+                    message="To enable fireshose based logging for WAF on CDN the deployment must be run from us-east-1"
```
ml019, line 514:
```suggestion
                    message="To enable firehose based logging for WAF on CDN the deployment must be run from us-east-1"
```

ml019 (changes requested):

PR #544, opened 2022-04-08
fix(backupstore): tag based conditions

roleyfoley:
## Intent of Change

## Description

- Adds the required resource tag to selection conditions
- Fixes the tag value used to align with our current standard tags

## Motivation and Context

Deployment fails when using tag based conditions as the values aren't quite right. 
Cloudformation doco doesn't reference the format but an error is raised and it is outlined in 
https://docs.aws.amazon.com/aws-backup/latest/devguide/assigning-resources.html 

## How Has This Been Tested?

Tested locally and added test suite

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/backupstore/setup.ftl
+++ b/aws/components/backupstore/setup.ftl
@@ -181,20 +181,26 @@
             [#-- Check if store matches are required --]
             [#local conditions = {} ]
             [#if isPresent(subSolution.Conditions.MatchesStore) ]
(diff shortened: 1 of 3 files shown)


ml019 (approved):
Weird that the ListOfTags doesn't use the same format as conditions.

PR #552, opened 2022-04-22
fix: log and account processing

roleyfoley:
## Intent of Change

## Description

- Changes the order for account level units to run iam before cmk
- Add support for CloudWatch Log resource policies

## Motivation and Context

Fixes issues outlined in https://github.com/hamlet-io/engine/pull/1973 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/inputseeders/aws/masterdata.json
+++ b/aws/inputseeders/aws/masterdata.json
@@ -1356,17 +1356,17 @@
             "Priority": 10,
             "DistrictType" : "account",
             "ResourceSets": {
+                "iam": {
+                    "deployment:Unit": "iam",
+                    "deployment:Priority" : 10,
+                    "ResourceLabels": [
+                        "iam"
+                    ]
+                },
                 "cmk" : {
                     "deployment:Unit" : "cmk",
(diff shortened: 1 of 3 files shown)


aws/inputseeders/aws/masterdata.json
```diff
@@ -1356,18 +1356,18 @@
...
-                },
                 "iam": {
                     "deployment:Unit": "iam",
-                    "deployment:Priority" : 30,
+                    "deployment:Priority" : 10,
                     "ResourceLabels": [
                         "iam"
                     ]
                 },
+                "cmk" : {
```
ml019, line 1366: We have traditionally created the cmk as the first thing for a segment. What drove the switch in ordering?

roleyfoley: The CMK grants access to ServiceRoles, in order to do that the service roles have to exist or the CMK will fail its creation. 

Its weird a resource policy that actually checks that the principals exist that you are trying to use

PR #563, opened 2022-05-04
feat(task): add ses smtp password generation task

roleyfoley:
## Intent of Change

## Description

- Defines a task that generates the required Sig4 signed password used by the SES smtp service

## Motivation and Context

The new password required for SES smtp services is a hash encoded sig4 signature. This is much more complex than the original approach and AWS guidance requires the use of a script like python to handle this. 

## How Has This Been Tested?

Tested locally with cli.

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/tasks/aws_ses_smtp_password/id.ftl
+++ b/aws/tasks/aws_ses_smtp_password/id.ftl
@@ -17,7 +17,7 @@
         },
         {
             "Names" : "AWSSecretAccessKey",
-            "Description" : "The secret key of the access key pair that will be used to send emails via SMPT",
+            "Description" : "The secret key of the access key pair that will be used to send emails via SMTP",
             "Types" : STRING_TYPE,
             "Mandatory" : true
         }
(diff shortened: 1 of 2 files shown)


aws/tasks/aws_ses_smtp_password/id.ftl
```diff
@@ -0,0 +1,25 @@
...
+    attributes=[
+        {
+            "Names" : "SESRegion",
+            "Description" : "The name of the region that the SMTP endpoint will be used",
+            "Types" : STRING_TYPE,
+            "Mandatory" : true
+        },
+        {
+            "Names" : "AWSSecretAccessKey",
+            "Description" : "The secret key of the access key pair that will be used to send emails via SMPT",
```
ml019, line 20:
```suggestion
            "Description" : "The secret key of the access key pair that will be used to send emails via SMTP",
```

ml019 (changes requested):

ml019 (approved):

PR #569, opened 2022-05-11
feat(rds): event support

ken9ross:
## Intent of Change

## Description
Enable RDS Events via Role: publish links from RDS to SNS topic

A Role of "publish" will establish all events to the SNS topic. An optional comma delimited list of events can follow publish after a space. Interested in feedback as to whether there is a better solution to this

## Motivation and Context
Fixes https://github.com/hamlet-io/engine/issues/140

## How Has This Been Tested?
Locally, deployed to AWS, testcases established

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/1991

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/services/rds/resource.ftl
+++ b/aws/services/rds/resource.ftl
@@ -22,6 +22,19 @@
     mappings=RDS_OUTPUT_MAPPINGS
 /]
 
+[#assign RDS_EVENT_OUTPUT_MAPPINGS =
+    {
+        REFERENCE_ATTRIBUTE_TYPE : {
+            "UseRef" : true
+        }
+    }
+]
+[@addOutputMapping
+    provider=AWS_PROVIDER
+    resourceType=AWS_RDS_EVENT_RESOURCE_TYPE
+    mappings=RDS_EVENT_OUTPUT_MAPPINGS
+/]
+
 [#assign RDS_CLUSTER_OUTPUT_MAPPINGS =
     {
         REFERENCE_ATTRIBUTE_TYPE : {
@@ -114,6 +127,8 @@
     performanceInsights
     performanceInsightsRetention
     tags
+    occurrence
+    links
     caCertificate
     engineVersion=""
     clusterMember=false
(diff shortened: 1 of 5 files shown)


aws/services/rds/resource.ftl
```diff
@@ -245,6 +260,49 @@
...
+                            "SourceIds" : [ getReference(id) ],
+                            "SourceType" : "db-instance"
+                        } +
+                        attributeIfContent(
+                            "EventCategories",
+                            linkRoles
+                        )
+                    outputs=RDS_EVENT_OUTPUT_MAPPINGS
+                /]
+            [/#if]
```
roleyfoley, line 303: I think this would be better to move under a dedicated parameter under the solution configuration. There is nowhere to currently document what the role values are supposed to be so this will be really hard to actually use. 

My suggestion would be to add a new root level attribute on the db component with something like this structure 


```
{
   "SystemNotifications" : {
          "badEvents" : {
                 "Categories" : [ "<a list of categories>" ],
                 "Links" : {
                     "bigYellingTopic" : { 
                              "Tier" : "mgmt",
                              "Component" : "alerts" 
                      }
                 }
             }
}
```

The categories attribute can then be a list of acceptable event categories, so they can be documented and maanged.

aws/services/rds/resource.ftl
```diff
@@ -324,4 +384,46 @@
...
+        [/#if]
+    [/#list]
```
roleyfoley, line 428: This shouldn't be part of a resource and should actually be in the component instead. The resource macros are for creating a particular resource and potentially be indpendent from the occurrence processing. 

This is also a lot duplication between the cluster and the instance, when it could be handled in the setup macro.

roleyfoley: Are there any permissions requirements? Usually there needs to be an inbound role on a component so the SNS topic's resource policy can be udpated.

roleyfoley (changes requested):

aws/services/rds/resource.ftl
ken9ross: I was assuming that the majority of the time, it would be all alerts to the topic (ie role = publish - with no following comma delimited list) with the occasional filtered set (only really when you needed to split the alerts into multiple topics, presumably for security restrictions). The construct you're proposing seems to be based on the assumption that the majority of the time it will be restricted alert sets.

Do you envision the majority of the cases being a restricted set or have I missed something?

aws/services/rds/resource.ftl
ken9ross: That will mean that most macros in resource will just be a call to cfResource won't it? I don't understand, but done

Permissions don't appear to be required from my testing

roleyfoley (approved):

PR #570, opened 2022-05-13
feat(alertslack): allow setting runtime with param

ken9ross:
## Intent of Change

## Description
Allow for lambda runtime to be set with a parameter for alertslack

## Motivation and Context
alertslack v1.2.0 has changes for python3.9 as the runtime. This is to allow it to be controlled via parameters

## How Has This Been Tested?
Locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/modules/cloudwatchslack/module.ftl
+++ b/aws/modules/cloudwatchslack/module.ftl
@@ -48,6 +48,16 @@
             "Default" : "8f194db4f6ed2b826387112df144f188451ba6db"
         },
         {
+            "Names" : "lambdaRuntime",
+            "Type" : STRING_TYPE,
+            "Description" : "runtime engine for lambda",
+            "Values" : [
+                "python3.6",
+                "python3.9"
+            ],
+            "Default" : "python3.6"
+        },
+        {
             "Names" : "namePrefix",
             "Type" : STRING_TYPE,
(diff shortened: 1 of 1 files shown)


aws/modules/cloudwatchslack/module.ftl
```diff
@@ -47,6 +47,16 @@
...
+            "Names" : "lambdaRuntime",
+            "Type" : STRING_TYPE,
+            "Description" : "runtime engine for lambda",
+            "Values" : [
+                "python3.6",
+                "python3.9"
+            ],
+            "Default" : "python3.6"
+        },
         {
```
roleyfoley, line 60: We should just set the lambda RunTime to Python3.9. The code doesn't have any new features in it and python3.9 is backwards compatible with python3.6 

There is no impact in changing runtime between these two and no need for choice.

roleyfoley (changes requested):

aws/modules/cloudwatchslack/module.ftl
ken9ross: Correct, but earlier versions of hamlet did not support above 3.8. Do all "in use" versions of hamlet support 3.9 (I'm particularly thinking out HA)

roleyfoley (approved):

PR #572, opened 2022-05-13
fix(ec2): fix ordering for cfn init commands

roleyfoley:
## Intent of Change

## Description

- Uses a list map to get the actual priority as a number

## Motivation and Context

When using CFN init scripts the ordering of tasks was determined by sorting on strings.  This meant that the ordering was based on same values rather than the number order of the priority. 

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (approved):
Funky use of maps!

PR #581, opened 2022-05-21
refactor(apigateway): authorization models

ml019:
## Intent of Change

## Description
Support additional model values for the case of IP filtering in combination with a lambda authorizer. Also rename the config attribute to more correctly reflect its purpose in controlling authorization rather than authentication.

## Motivation and Context
When used with a lambda authorizer, the default value of "IP" incorrectly provides an explicit ALLOW rather than relying on it to come from the policy provided by the authorizer. By providing explicit values to be used with the authorizer, the configuration can be validated as appropriate.

## How Has This Been Tested?
Local template generation

### Prerequisite PRs:
- https://github.com/hamlet-io/engine/pull/1995

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -281,7 +281,7 @@
                 [#break]
 
             [#case "SIG4ORIP" ]
-            [#case "SIG4_OR_IP" ]
+            [#case "aws:SIG4_OR_IP" ]
             [#case "AUTHORIZER_OR_IP" ]
             [#case "AUTHORISER_OR_IP" ]
                 [#-- Resource policy provides ALLOW on IP --]
@@ -300,7 +300,7 @@
                 [#break]
 
             [#case "SIG4ANDIP" ]
-            [#case "SIG4_AND_IP" ]
+            [#case "aws:SIG4_AND_IP" ]
             [#case "AUTHORIZER_AND_IP" ]
             [#case "AUTHORISER_AND_IP" ]
                 [#-- If IP doesn't match, EXPLICIT DENY regardless of SIG4 --]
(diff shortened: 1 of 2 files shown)


aws/components/apigateway/setup.ftl
```diff
@@ -224,17 +238,36 @@
...
+            [#local apiPolicyAuth = "AUTHORISER_AND_IP" ]
+        [/#if]
+
+        [#-- If cognito authorizers are in use, the AuthorisationModel must be IP --]
+        [#if cognitoPools?has_content && (apiPolicyAuth != "IP") ]
+            [@warn
+                message="Authorization model of \"" + apiPolicyAuth + "\" is not compatible with the use of cognito authorizers. Forcing to \"IP\"."
+            /]
+            [#local apiPolicyAuth = "IP" ]
+        [/#if]
```
roleyfoley, line 256: Should we make these fatal and have people change them. If the configuration option is saying it should be this value then we should probably have people change it in the config.

roleyfoley (changes requested):

aws/components/apigateway/setup.ftl
ml019: I guess that wouldn't be a breaking change in that it will fix a bug, but it will break previously "working" configurations. However it will only kick in with the update that fixes the bug. 

On balance I think you are right - I'll tweak them.

ml019:
@roleyfoley I've added prefixed values for the SIG4 models. In a future breaking change, we can remove the non-prefixed ones. I'll rise a ticket to add this to the list for the next breaking change.

roleyfoley (approved):

PR #600, opened 2022-06-07
fix(cert): raise error on invalid FQDN

ken9ross:
## Intent of Change

## Description

Raise non-stopping fatal error when api gateway domain name does not include at least one dot combines with more general https://github.com/hamlet-io/engine/pull/2008

## Motivation and Context

Fixes https://github.com/hamlet-io/engine/issues/1665

## How Has This Been Tested?

Couldn't see how to set up a failure case to test - any advice

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/2008

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/state.ftl
+++ b/aws/components/apigateway/state.ftl
@@ -298,24 +298,28 @@ created in either case.
     [#local customDomainResources = {} ]
     [#list customDomains as domain]
         [#local customFqdn = formatDomainName(customHostName, domain) ]
-        [#local customFqdnParts = splitDomainName(customFqdn) ]
-        [#local customFqdnId = formatResourceId(AWS_APIGATEWAY_DOMAIN_RESOURCE_TYPE, customFqdnParts) ]
-        [#local customDomainResources +=
-            {
-                customFqdn : {
-                    "domain" : {
-                        "Id" : customFqdnId,
-                        "Name" : customFqdn,
-                        "CertificateId" : certificateId,
-                        "Type" : AWS_APIGATEWAY_DOMAIN_RESOURCE_TYPE
-                    },
-                    "basepathmapping" : {
-                        "Id" : formatDependentResourceId(AWS_APIGATEWAY_BASEPATHMAPPING_RESOURCE_TYPE, customFqdnId),
-                        "Stage" : mappingStage,
-                        "Type" : AWS_APIGATEWAY_BASEPATHMAPPING_RESOURCE_TYPE
+        [#if !customFqdn?contains(".") ]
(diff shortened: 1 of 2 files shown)


aws/services/acm/id.ftl
```diff
@@ -32,6 +32,10 @@
 [#function formatDomainCertificateId certificateObject, hostName=""]
     [#local primaryDomain = getCertificatePrimaryDomain(certificateObject) ]
     [#if primaryDomain.Name?has_content ]
+        [#if !primaryDomain.Name?contains(".") ]
+            [@fatal message="Missing domain name for ACM primary domain" context=solution /]
```
roleyfoley, line 36: Doesn't look like solution is available at this point? 
Is it worth performing this validation in getCertificatePrimaryDomain instead as that is more where the issue is occurring?

aws/components/apigateway/state.ftl
```diff
@@ -298,24 +298,28 @@ created in either case.
...
-                        "CertificateId" : certificateId,
-                        "Type" : AWS_APIGATEWAY_DOMAIN_RESOURCE_TYPE
-                    },
-                    "basepathmapping" : {
-                        "Id" : formatDependentResourceId(AWS_APIGATEWAY_BASEPATHMAPPING_RESOURCE_TYPE, customFqdnId),
-                        "Stage" : mappingStage,
-                        "Type" : AWS_APIGATEWAY_BASEPATHMAPPING_RESOURCE_TYPE
+        [#if !customFqdn?contains(".") ]
+            [@fatal message="Missing domain name for API gateway" context=solution /]
+        [#else]
```
roleyfoley, line 303: Removing the state will cause issues in the setup routine when it is run. This is still run before an alert is generated. If the state is depdended on in the setup process we won't see this alert.

roleyfoley (changes requested):

aws/components/apigateway/state.ftl
ken9ross: Since the fatal error doesn't have a stop on it and I'm just excluding the customFqdn doesn't it continue to process just without the custom domain; and the state and setup seems to handle a missing customDomains already. What am I missing?

aws/services/acm/id.ftl
ken9ross: Moved to getCertificatePrimaryDomain

roleyfoley (approved):

ml019:
Merging without branch update per approval from @roleyfoley

PR #602, opened 2022-06-10
feat(lb): add support for enabling/disabling conditions

roleyfoley:
## Intent of Change

## Description

-Don't include conditions which aren't enabled 

## Motivation and Context

Implementation of https://github.com/hamlet-io/engine/pull/2011

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/2011

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/lb/setup.ftl
+++ b/aws/components/lb/setup.ftl
@@ -567,44 +567,46 @@
             [/#if]
 
             [#list solution.Conditions as id, condition ]
-                [#switch condition.Type ]
-                    [#case "httpHeader" ]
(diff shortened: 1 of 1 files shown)


ml019 (approved):
We've done this elsewhere as a filter. Should we have an idiomatic preference? Approving for now as functionally correct.

roleyfoley:
Filters only work on sequences not hashes. This one uses the key of the hash as the Id and its used inside of the loop. 
So I would say yes filter where possible, but sometimes not always possible

PR #611, opened 2022-06-22
feat(image): aws support for the image component

roleyfoley:
## Intent of Change

## Description

- Implements the image component for the aws provider
- Adds service/resource support for ECR Repositories

## Motivation and Context

This isolates images into their own environments and allows for image management to be completely managed via the engine 
ECR repositories for AWS allow for greater control for docker based images and consolidate the images into their own repository

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (commented):
We discussed previously the idea of a local staging area distinct form the source of the image. So deployment templates would drive off the staging store but pull from the source if the image wasn't available locally. I'm thinking of things like the claims system authorizer where it would be built in the claims system registry (or maybe a tenant level registry), and then pulled into the staging store to ensure deployments in accounts can operate independently and to avoid cross account policies etc. 

I'm assuming what is here thus far just gives us what we have now, and consumers would just drive off the attributes it offers so in future, we could cache and then change what attribute values the component provides?

roleyfoley:
Intention with the way I did it was to align the concept that we apply across everything else that districts are essentially isolated from one another. 

Under the model documented here the sharing/transferring/caching whatever is outside of the case of deployment tooling and could be managed outside of our deployment tooling.

ml019 (approved):

PR #617, opened 2022-06-27
refactor!: remove use of segmentObject

roleyfoley:
## Intent of Change

## Description

- Moves to using the more generic getTiers methods which supports occurrences in different district types
- Removes the use of the AppPublicDataPrefix - This has been removed from the engine to reduce the complexity in managing these different prefixes and is also considered a security risk. This was removed from the core engine to save having to migrate the segment level configuration required to use the public data prefix in the first place.

The preference for public data would be to use an explicitly defined S3 bucket for this particular purpose. It makes it easier to understand from a user perspective and reduces panic reporting when a core S3 bucket has some limited public access

## Motivation and Context

Aligns with the change in the engine to support occurrence based deployments across multiple district types instead of just the segment base districts

## How Has This Been Tested?

tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- https://github.com/hamlet-io/engine/pull/1968

### Consumer Actions:

- None

### DIFF
--- a/aws/components/network/state.ftl
+++ b/aws/components/network/state.ftl
@@ -32,8 +32,7 @@
 
     [#local subnets = {} ]
     [#-- Define subnets --]
-    [#list segmentObject.Network.Tiers.Order as tierId]
-        [#local networkTier = getTier(tierId) ]
+    [#list getTiers()?filter(x -> x.Network.Enabled && x.Active ) as networkTier]
         [#if ! (networkTier?has_content && networkTier.Network.Enabled &&
                     networkTier.Network.Link.Tier == core.Tier.Id && networkTier.Network.Link.Component == core.Component.Id &&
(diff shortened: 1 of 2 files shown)


ml019:
@roleyfoley not sure I quite understand the removal of the appdata public stuff. What was the scenario we introduced it for, and what will be the approach if this change is merged?

roleyfoley:
> @roleyfoley not sure I quite understand the removal of the appdata public stuff. What was the scenario we introduced it for, and what will be the approach if this change is merged?

Forgot to mention that this was part of the solution level update that removes it. Was trying to reduce the usage of the segment configuration which the changes to networking and the tier structure help with

ml019:
OK - I see you've updated the PR comment and the proposal makes sense. If you want a public bucket, you need to explicitly create one and set the bucket policy to allow public access.

PR #619, opened 2022-07-09
fix: support dns zone creation without network

roleyfoley:
## Intent of Change

## Description

Updates state support in dnszone to align with what is current available

## Motivation and Context

- Allows for creating a DNS zone without the network profile
- Supports creating the zone then adding the network later

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/dnszone/state.ftl
+++ b/aws/components/dnszone/state.ftl
@@ -20,19 +20,52 @@
     [#local domainName = getCertificatePrimaryDomain(domainObject).Name ]
     [#local zoneId = formatResourceId(AWS_ROUTE53_HOSTED_ZONE_RESOURCE_TYPE, core.Id) ]
 
+    [#local attributes = {}]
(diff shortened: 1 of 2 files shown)


aws/components/dnszone/state.ftl
```diff
@@ -5,76 +5,38 @@
...
-                "Type": AWS_ROUTE53_HOSTED_ZONE_RESOURCE_TYPE
-            }
-        }]
-
-        [#local attributes = {
-            "DOMAIN" : domainName,
-            "ZONE" : getExistingReference(zoneId)
-        }]
-
-    [#else]
```
ml019, line 47: The ability to manually configure the zone from external info seems to have been removed. This wasn't mentioned in the description,

The aim of this was to permit a pre-existing zone to be imported.

PR #622, opened 2022-08-02
fix(computetask): windows directory creation

roleyfoley:
## Intent of Change

## Description

- Use the powershell new-item command with -Force to handle re-running the cfn-init commands without failure
- Removes the initial directory setup script for windows as it only creates one directory that we no longer need

## Motivation and Context

Noticed during re-runs of cfn-init commands that the directory creation was raising errors. This uses force to create the directory which will handle the existing directory. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
@roleyfoley I see an init section was removed as well but not mentioned in the description. Is it no longer needed?

roleyfoley:
> @roleyfoley I see an init section was removed as well but not mentioned in the description. Is it no longer needed?

Yep it was creating a single directory that we don't need. Not really sure why it was implemented in such a special way.

PR #625, opened 2022-08-13
feat(cloudfront): complex cdn scenarios

roleyfoley:
## Intent of Change

## Description

- Adds support for Cloudfront CachePolicies, Origin Request policies which split cache handling from origin requests to fine tune how the cache is managed
- Adds support for Cloudfront Response Policies which support injecting headers sent back to the client, this includes predefined support for HSTS, CORS and other security protocols
- Adds support for the new subcomponents origin, cache policy and response policy which can be used in complex cdn scenarios to reduce resource usage and complexity

## Motivation and Context

- Cache and Origin request policies were [introduced](https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-cloudfront-announces-cache-and-origin-request-policies/) to provide greater control over what is included in the cache key used by cloudfront vs what is sent to the origin. This is useful in scenearios where you need to forward headers or cookies for a session, but the content returned is the same across all users, so in this case you don't want to cache based on cookies and but they need to be sent to the origin when the cache expires
- Splitting out the origin cache policy and response policies into separate ( origin and cache policy are optional) reduces the complexity in the resource that is deployed. There is a hard limit on all cloudfront policies to 20 per account, so having duplicated policies on complex CDN routes gets tricky 
- Adding support for [Response Header Policies](https://aws.amazon.com/blogs/networking-and-content-delivery/amazon-cloudfront-introduces-response-headers-policies/) allows the CDN to automatically manage standard security headers like CORS, HSTS etc without needing Lambda@Edge functions all that jazz.

## How Has This Been Tested?

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/2029

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/apigateway/setup.ftl
+++ b/aws/components/apigateway/setup.ftl
@@ -719,6 +719,43 @@
 
         [#-- Create a CloudFront distribution if required --]
         [#if cfResources?has_content]
+
+            [#local cachePolicy = cfResources["cachePolicy"] ]
+            [#local originRequestPolicy = cfResources["originRequestPolicy"]]
+
+            [@createCFCachePolicy
+                id=cachePolicy.Id
+                name=cachePolicy.Name
+                ttl={
+                    "Min": 0,
+                    "Max": 0,
+                    "Default": 0
+                }
+                headerNames=[] cookieNames=[]
+                queryStringNames=[] compressionProtocols=[]
+            /]
+
+            [@createCFOriginRequestPolicy?with_args(
+                getOriginRequestPolicy(
+                    "originRequestPolicy",
+                    "LinkType",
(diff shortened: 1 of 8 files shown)


ml019:
@roleyfoley looks like the test solution needs to be tweaked a bit? Are there changes to solutions that are needed for this change? We should class it as breaking if so ...

roleyfoley:
> @roleyfoley looks like the test solution needs to be tweaked a bit? Are there changes to solutions that are needed for this change? We should class it as breaking if so ...

Testing fixed up. There is one minor configuration change if you try to use the existing cache policy config on a route that it is not the default, you will be told to move to the new cache policy setup.

PR #654, opened 2022-10-06
feat(waf): add support for more wafv2 rules

roleyfoley:
## Intent of Change

## Description

- Adds support for defining conditional rate based rules in wafv2
- Adds support for using Vendor Managed Rule groups which include detailed rule management from AWS
- Both are controlled as rules as they can have conditions applied to override their usage

## Motivation and Context

AWS Implementation of https://github.com/hamlet-io/engine/pull/2055 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/services/waf/resource.ftl
+++ b/aws/services/waf/resource.ftl
@@ -60,6 +60,7 @@
     [#local v2Statements = []]
     [#list conditions as condition]
         [#local v2WkStatement = []]
+
         [#list condition.Filters as filter]
             [#switch condition.Type]
                 [#case AWS_WAF_SQL_INJECTION_MATCH_CONDITION_TYPE]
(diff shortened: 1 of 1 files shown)


aws/services/waf/resource.ftl
```diff
@@ -351,6 +360,20 @@
                     version=version /]
             [/#if]
             [#switch version]
+
+                [#switch rule.Engine ]
```
ml019, line 364: Not sure this will work as the code isn't inside a case block for the switch based on version?

ml019 (changes requested):

aws/services/waf/resource.ftl
roleyfoley: Had it locally but didn't push.

ml019 (approved):

PR #655, opened 2022-10-06
fix(cdn): use cloudformation to find cdn id

roleyfoley:
## Intent of Change

## Description

- Use the Cloudformation stack output to determine the CF ID when running an invalidation

## Motivation and Context

When a CF Id has changed and not been reflected in the CMDB the invalidation command doesn't work as it tries to invalidate the wrong ID. Using the CFN stack ensures that its up to date. 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/cdn/setup.ftl
+++ b/aws/components/cdn/setup.ftl
@@ -833,16 +833,14 @@
         [#if invalidationPaths?has_content && getExistingReference(cfId)?has_content && (wafVersion == "v1") ]
             [@addToDefaultBashScriptOutput
                 [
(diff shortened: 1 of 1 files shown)


roleyfoley:
> @roleyfoley when does the cfid change? change is fine but wondering what the scenario is?

I'm not too sure how it happened, its rare that the cfId will change, but I was getting it.

ml019 (approved):
Weird but ok - agree this is more robust

PR #668, opened 2022-10-21
fix(waf): handle WAFProfiles and metric name

roleyfoley:
## Intent of Change

## Description

- Use the WAF Profile from the security profile at all times
- Fix the name for the default rule metric name

## Motivation and Context

- The WAFProfile was being ignored if OWASP wasn't enabled on the WAF configuration
- The default metric would change to the name of the last rule or fail when no rules are present

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/services/waf/resource.ftl
+++ b/aws/services/waf/resource.ftl
@@ -536,11 +536,10 @@
         [#break]
 
         [#case "v2"]
-            [#local defAction = defaultAction?lower_case?cap_first ]
             [#local properties=
                 {
                     "DefaultAction" : {
-                        defAction: {}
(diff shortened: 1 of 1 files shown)


ml019:
Wondering if we should add a separate entry in the security profile for the OWASP version rather than hard code?

roleyfoley:
I think the OWASP option should really be removed as you can achieve the same result with the SecurityProfile configuration option. Given profiles are used more broadly across our components I think this is the more hamlet way of going.

ml019 (approved):

PR #669, opened 2022-10-21
fix(user): pseudo stack handling and ses removal

roleyfoley:
## Intent of Change

- Breaking change (fix or feature that would cause existing functionality to change)

## Description

- Ensures a unique psuedo stack is generated for each user instead of overriding when there are multiple users in a deploymentUnit
- Remove the SES password generation as the method being used is no longer accepted by SES API endpoints and instead requires Sig4 signing
- Cleans up some of the formatting for generating the bash script

## Motivation and Context

When creating a deployment that has multiple users the pseudo stack for credentials would be overwrriten for each user created 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- The SES User password has been removed and replaced with a runbook/module as password generation is now dependent on the region of SES you are using. 

The Module https://github.com/hamlet-io/engine-plugin-aws/blob/master/aws/modules/ses_smtp_user/module.ftl can be used to create an SMTP user in AWS along with a runbook to generate the username and password based on a provided region.

### DIFF
--- a/aws/components/user/setup.ftl
+++ b/aws/components/user/setup.ftl
@@ -258,8 +258,6 @@
     [/#if]
 
     [#if deploymentSubsetRequired("epilogue", false)]
-
-        [#local credentialsPseudoStackFile = "\"$\{CF_DIR}/$(fileBase \"$\{BASH_SOURCE}\")-credentials-pseudo-stack.json\"" ]
         [@addToDefaultBashScriptOutput
             content=
             [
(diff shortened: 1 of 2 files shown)


ml019:
I don't think anything was using the SES password? Did you have an example? Just thinking about the breaking change aspect of this.

roleyfoley:
An example of which? the use of the SES Password or the module?

ml019:
Use of the SES Password

ml019 (approved):

PR #710, opened 2023-03-21
refactor(mta): move to using cfn for config set event destination

roleyfoley:
## Intent of Change

## Description

- Moves from CLI to cloudformation resource for adding the configuration set event destination in the MTA component.
- Adds the MAIL_DOMAIN as an attribute so that it can be used by components to know their expected domain

## Motivation and Context

Removes the use of the CLI and allows for streamlined management of the mta component through cfn

## How Has This Been Tested?

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/mta/setup.ftl
+++ b/aws/components/mta/setup.ftl
@@ -79,8 +79,6 @@
 
                 [#switch solution.Action]
                     [#case "log"]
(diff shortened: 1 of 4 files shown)


ml019 (approved):
Looks like CF has finally caught up. Nice when it is all in the template

PR #711, opened 2023-03-21
feat: cmk access and SQS SSE

roleyfoley:
## Intent of Change

## Description

- Relax the conditions for ses access to the cmk so that the same policy can be used for send and receive logging. Currently the send logging doesn't work with encryption as the context vars used in the policy aren't included for send events
- Adds support for sqs to access the CMK when using SSE
- Adds the required configuration to enable SSE on SQS queues using Customer managed keys
- SSE-CMK is required to use an encrypted queue with SNS topics and also aligns with our use of encryption at rest across other services
- Adds an enabled by default policy to only allow HTTPS based connections to the SQS queue similar to what is available in S3

## Motivation and Context

This aligns with the encouraged best practices of ensuring that encryption is used to store all data as it is hosted in systems and passed from place to place

## How Has This Been Tested?

Tested locally and with deployment

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019 (commented):
There is some stuff around ECS changes and docker volumes which doesn't appear part of the scope of this PR?

roleyfoley:
> There is some stuff around ECS changes and docker volumes which doesn't appear part of the scope of this PR?

Fixed up

ml019 (approved):

PR #737, opened 2023-05-17
feat(network): support default sg nacl control

roleyfoley:
## Intent of Change

## Description

- Adds support for setting rules on the default security group and NACL on a VPC created by hamlet
- Allows for passing a Ref through to a security group and NACL rule
- Fixes the port used for outbound rules in return traffic

## Motivation and Context

AWS VPC's always include a default Security group and NACL which cannot be removed, being able to set base rules on these resources ensures that they can't be used for the wrong purpose

## How Has This Been Tested?

Tested locally 

### Prerequisite PRs:

- https://github.com/hamlet-io/engine/pull/2107

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/network/setup.ftl
+++ b/aws/components/network/setup.ftl
@@ -457,14 +457,6 @@
 
             [#if deploymentSubsetRequired(NETWORK_COMPONENT_TYPE, true)]
 
-                [#if ! resources["networkACL"].DefaultACL ]
-                    [@createNetworkACL
-                        id=networkACLId
-                        vpcId=vpcResourceId
-                        tags=getOccurrenceTags(occurrence)
-                    /]
-                [/#if]
-
                 [#if resources["networkACL"].DefaultACL ]
                     [#local networkACLId = {
                         "Fn::GetAtt": [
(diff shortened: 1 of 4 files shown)


aws/components/network/setup.ftl
```diff
@@ -418,11 +456,23 @@
...
-                    tags=getOccurrenceTags(occurrence)
-                /]
+
+                [#if ! resources["networkACL"].DefaultACL ]
+                    [@createNetworkACL
+                        id=networkACLId
+                        vpcId=vpcResourceId
+                        tags=getOccurrenceTags(occurrence)
+                    /]
+                [/#if]
```
ml019, line 466: should this be an else?

aws/components/network/setup.ftl
```diff
@@ -461,7 +511,7 @@
                         [#local forwardIpAddresses = getGroupCIDRs(ruleConfig.Source.IPAddressGroups, true, occurrence)]
                         [#local forwardPort = (ports[ruleConfig.Destination.Port])!{}]
                         [#local returnIpAddresses = [ "0.0.0.0/0" ]]
-                        [#local returnPort = (ports[ruleConfig.Destination.Port])!{}]
+                        [#local returnPort = (ports[ruleConfig.Source.Port])!{}]
```
ml019, line 514: Was this a bug? Looks like a significant change not talked about in the description?

roleyfoley: - Fixes the port used for outbound rules in return traffic 

I don't think we've ever actually applied NACLs that this would have been noticed on.

aws/components/network/setup.ftl
roleyfoley: Fixed

ml019 (approved):

PR #738, opened 2023-05-17
feat(waf): add support for regex pattern set and custom responses

roleyfoley:
## Intent of Change

## Description

- Adds support for creating rules using the regex pattern set resource and then being able to reference it in a rule
- Adds support for custom block responses ( https://docs.aws.amazon.com/waf/latest/developerguide/customizing-the-response-for-blocked-requests.html ) to control how clients find out about the block that occurred

## Motivation and Context

- This was partially implemented in the migration to wafv2 but a couple of the steps seem to have been missed along the way 
- Custom block responses provide a way to respond to clients in the same format that they would expect for the application or provide them with a way to get in touch if this isn't expected behaviour

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/services/waf/resource.ftl
+++ b/aws/services/waf/resource.ftl
@@ -221,7 +221,7 @@
                             "Scope" : regional?then("REGIONAL","CLOUDFRONT"),
                             "IPAddressVersion" : "IPV4",
                             "Addresses": asFlattenedArray(
-                                filters?map(
+                                filters?map(filter ->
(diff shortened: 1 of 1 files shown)


ml019:
Think this has a dependency on https://github.com/hamlet-io/engine/pull/2108?

ml019 (approved):

PR #755, opened 2023-06-14
fix(baseline): permissions for logging

roleyfoley:
## Intent of Change

## Description

- Only allow s3:PutObject instead of s3:PutObject*
- Add support for the new region lb logging to use the fixed principal instead of a fixed AWS account

## Motivation and Context

- AWS Security hub benchmark recommends only allowing minimal explicit actions when adding bucket policy for external accounts. Since the LB log forwarded uses an external account the reporting trips up on the current policy 
- Adding the new regions allows for using the new regions if we need to 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/baseline/setup.ftl
+++ b/aws/components/baseline/setup.ftl
@@ -353,13 +353,34 @@
                     [#case "operations" ]
 
                         [#local bucketPolicy +=
-                            [#-- Permission for ELB to write access logs to s3 --]
-                            s3WritePermission(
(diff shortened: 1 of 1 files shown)


ml019 (approved):
I'll approve this one to get it into the unicycle. I'll rebase my PR to switch to extensions.

PR #771, opened 2023-07-25
refactor: remove use of component ids

roleyfoley:
## Intent of Change

## Description

- Removes the use of component ids from services that are still using them
- Removes Id formatting functions which can be handled easily using formatResourceId
- Add migration for ECS ec2 launch config from component Id to resource Id
- Add migration for Ec2 security groups to use occurrence Id
- Remove product level bucket Id naming as we don't support product level deployments

## Motivation and Context

Component Ids don't include Version or Instance components in the Id so if you have multiple instances then some resources will overlap between the two, notably this currently happens on Ec2 Instance Profiles and Roles. 
When this happen the get reference process will use the profile/role from the other instance and the permissions don't line up as expected. 

This change will replace the instanceprofile and Role attached to an ec2 Instance, but both don't require any replacement actions during their change over. 

Ec2 Launch Configs for ECS are also affected but to reduce the impact they will only be changed to the new Id structure if it is a new deployment of ECS not an existing one which uses the older naming format

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None


ml019:
A default template for solution lg units was removed as well. I'm assuming that was a fallback template which never ended up being used?

roleyfoley:
> A default template for solution lg units was removed as well. I'm assuming that was a fallback template which never ended up being used?

Yeah I couldn't figure out what it did and everything should have their own log groups to use anyway.

ml019 (approved):

PR #786, opened 2023-10-13
feat(topic): add fixed endpoint subscriptions

roleyfoley:
## Intent of Change

## Description

AWS implementation of https://github.com/hamlet-io/engine/pull/2127

## Motivation and Context

Make it easier to setup subscriptions 

## How Has This Been Tested?

Tested locally

### Prerequisite PRs:

- None

### Dependent PRs:

- None

### Consumer Actions:

- None

### DIFF
--- a/aws/components/topic/setup.ftl
+++ b/aws/components/topic/setup.ftl
@@ -247,6 +247,28 @@
                         /]
                     [/#if]
                 [/#list]
+
+                [#list solution.Endpoints as id, endpoint]
+                    [#switch endpoint.Protocol ]
+                        [#case "http"]
+                        [#case "https" ]
+                            [#local deliveryPolicy = getSNSDeliveryPolicy(solution.DeliveryPolicy) ]
+                            [#break]
+                    [/#switch]
+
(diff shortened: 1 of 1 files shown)


aws/components/topic/setup.ftl
```diff
@@ -247,6 +247,28 @@
                         /]
                     [/#if]
                 [/#list]
+
+                [#list solution.Endpoints as id, endpoint]
+                    [#switch endpoint.Protocol ]
+                        [#case "http"]
+                        [#case "https" ]
+                            [#local deliveryPolicy = getSNSDeliveryPolicy(solution.DeliveryPolicy) ]
```
ml019, line 255: Think this needs to be initialised as part of the loop for the case of the endpoint not being http/s.

ml019 (changes requested):

aws/components/topic/setup.ftl
roleyfoley: Ta it was falling back to the one defined for the links

